PR #3, opened 2020-06-22
Use docker:// scheme for installing docker images

### LINKED ISSUE #1: Add hub.docker.com as a provider
# Description
CLIs are sometimes provided as docker images and allow users to use the CLI without having to worry about installing the binary correct for their architecture.

# Usability
Users of bin should be able to install docker images that represent CLIs if they were a regular binary that needs to be downloaded:
```
bin install hub.docker.com/r/hashicorp/terraform
terraform init
```

matipan:
# Description
Docker images live in many different registries, bin needs to support downloading from all of them. For starter, we need to differentiate between github, docker or any other provider. To do this, when trying to install docker images users need to specify the `docker://` scheme:
```
bin install docker://postgres
```
We also need to support tags so that users can do:
```
bin install docker://hashicorp/terraform:light
```

Closes #1

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -149,19 +149,6 @@ func saveToDisk(f *providers.File, path string, overwrite bool) error {
 
 	var outputFile = io.MultiReader(&buf, f.Data)
 
-	if t != matchers.TypeElf && t != matchers.TypeGz {
-		// if its not elf or gz then its an executable text file
-		// that needs to be written as is
-		file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0766)
-		if err != nil {
-			return err
-		}
-		defer file.Close()
-
-		_, err = io.Copy(file, outputFile)
-		return err
-	}
-
 	if t == matchers.TypeGz {
 		fileName, file, err := processTarGz(outputFile)
 		if err != nil {
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -137,23 +137,30 @@ func getFinalPath(path, fileName string) (string, error) {
 //TODO check if other binary has the same hash and warn about it.
 //TODO if the file is zipped, tared, whatever then extract it
 func saveToDisk(f *providers.File, path string, overwrite bool) error {
-
 	defer f.Data.Close()
 
 	var buf bytes.Buffer
 	tee := io.TeeReader(f.Data, &buf)
 
 	t, err := filetype.MatchReader(tee)
-
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


cmd/install.go
```diff
@@ -137,23 +137,30 @@ func getFinalPath(path, fileName string) (string, error) {
...
 	}
 
+	var outputFile = io.MultiReader(&buf, f.Data)
+
 	if t != matchers.TypeElf && t != matchers.TypeGz {
-		return fmt.Errorf("File type [%v] not supported", t)
-	}
+		// if its not elf or gz then its an executable text file
+		// that needs to be written as is
+		file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0766)
```
marcosnils, line 155: Hmm why not using variables here and follow the current function flow execution. It's a bit hard for me to understand that this is an alternative "valid" flow that this function can have.

pkg/providers/docker.go
```diff
@@ -6,25 +6,25 @@ import (
...
 
+	distreference "github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/reference"
 	"github.com/moby/moby/client"
 )
 
 const (
-	sh = `docker run --rm -i -t -v ${PWD}:/tmp/cmd -w /tmp/cmd %s/%s:%s "$@"`
+	sh = `docker run --rm -i -t -v ${PWD}:/tmp/cmd -w /tmp/cmd %s:%s "$@"`
```
marcosnils, line 18: //TODO not sure if this will work on windows. Can we add a TODO here so we don't forget about it?

pkg/providers/providers.go
```diff
@@ -41,9 +47,5 @@ func New(u string) (Provider, error) {
 		return newGitHub(purl)
 	}
 
-	if strings.Contains(purl.Host, "hub.docker.com") || strings.Contains(purl.Host, "docker.io") {
```
marcosnils, line 44: Before this I could do `bin install github.com/franela/matipan`. Can we fallback to `https://` seems like this won't work anymore now. Can we add a case for that there?

pkg/providers/docker.go
matipan: Probably won't. Unless they are on the WSL. I'll add the TODO 👍

marcosnils: Makes sense. I can test this and adapt after merging.

cmd/install.go
matipan: Fixed!

marcosnils:
:tada:

PR #13, opened 2020-07-23
Try OS || Arch if OS && Arch yields no matches

### LINKED ISSUE #12: Filtering on OS && Arch matches fewer releases
https://github.com/marcosnils/bin/blob/68f4df8f30c611b2fbbfa4d22481eecdb91d1c55/pkg/providers/github.go#L48 test for the presence of both the OS and the architecture. Some repos only list one of them, e.g. the OS, or presume the OS and list the archs.

sirlatrom:
Fixes #12.

Specifically matches https://github.com/stedolan/jq/releases/tag/jq-1.6.

### DIFF
--- a/pkg/providers/github.go
+++ b/pkg/providers/github.go
@@ -49,6 +49,14 @@ func filterAssets(as []*github.ReleaseAsset) (*githubFileInfo, error) {
 			matches = append(matches, &githubFileInfo{a.GetBrowserDownloadURL(), a.GetName()})
 		}
 	}
+	if len(matches) == 0 {
+		for _, a := range as {
+			lowerName := strings.ToLower(*a.Name)
+			if bstrings.ContainsAny(lowerName, config.GetOS()) || bstrings.ContainsAny(lowerName, config.GetArch()) {
+				matches = append(matches, &githubFileInfo{a.GetBrowserDownloadURL(), a.GetName()})
+			}
+		}
+	}
 
 	var gf *githubFileInfo
 	if len(matches) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/providers/github.go
```diff
@@ -49,6 +49,14 @@ func filterAssets(as []*github.ReleaseAsset) (*githubFileInfo, error) {
 			matches = append(matches, &githubFileInfo{a.GetBrowserDownloadURL(), a.GetName()})
 		}
 	}
+	if len(matches) == 0 {
```
marcosnils, line 52: Hmm this is a big dangerous.. since you could potentially be downloading a binary that's not for your OS or Arch..

Did you add this tying the `jq` case? I believe in that case it's not necessary since jq has both arch and os in the filename.. 

Not sure if I like this approach since it's very likely that you could end up downloading the incorrect binary. 

If someone publishes something like `jq-linux` I don't think we should support it since that could be any platform.

pkg/providers/github.go
```diff
@@ -143,6 +151,14 @@ func (g *gitHub) Fetch() (*File, error) {
 func sanitizeName(name, version string) string {
 	name = strings.ToLower(name)
 	replacements := []string{}
+
+	for _, osName := range config.GetOS() {
```
marcosnils, line 155: Since we're doing a nested loop here, we could move this logic inside here https://github.com/marcosnils/bin/pull/13/files#diff-d73633b9b2f88a927030f7bd9c67fccfR162-R170 so we don't loop again afterwards.

pkg/providers/github.go
sirlatrom: If only one asset matches either OS or Arch then it's the best bet. If several match _either_ OS _or_ Arch then the user currently gets an ordered list to choose from.

pkg/providers/github.go
sirlatrom: The most important part is that we'd check the `_`/`-` + `osName` + `archName` combo first.

pkg/providers/github.go
sirlatrom: > If someone publishes something like `jq-linux` I don't think we should support it since that could be any platform.

PS: Yes that is true, but if it is the only match and it happens to be correct, then isn't it better to err on the side of getting a binary?

PR #15, opened 2020-07-29
Reorder replacements to match osName+archName first

sirlatrom:
db7ff1addb3fdd975e63f7ce23fff57a703d3ece introduced a regression wrt. #12 (PR #13). This reordering ensures the combination of OS + Arch without delimiters is matched first.

### DIFF
--- a/pkg/providers/github.go
+++ b/pkg/providers/github.go
@@ -180,8 +180,6 @@ func sanitizeName(name, version string) string {
 	// generate the replacements? IDK.
 	firstPass := true
 	for _, osName := range config.GetOS() {
-		replacements = append(replacements, "_"+osName, "")
-		replacements = append(replacements, "-"+osName, "")
 		for _, archName := range config.GetArch() {
 			replacements = append(replacements, "_"+osName+archName, "")
 			replacements = append(replacements, "-"+osName+archName, "")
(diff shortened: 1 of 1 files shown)


marcosnils:
@sirlatrom thx for this :pray: . Since we have a test for the sanitizeName function (it's the only test :joy:). Can we add a case for that?

sirlatrom:
> 
> 
> @sirlatrom thx for this 🙏 . Since we have a test for the sanitizeName function (it's the only test 😂). Can we add a case for that? We should also rebase now since we merged #16

I'll rebase this and stand by to rebase #17 afterwards.

sirlatrom:
Actually come to think of it this one should just be closed as the fix was included as part of #16.

marcosnils:
> Actually come to think of it this one should just be closed as the fix was included as part of #16.

I didn't see that. You fooled me with the "refactoring" title :). Can we add a test case for this? Since we already have the Sanitize test case?

sirlatrom:
> 
> 
> > Actually come to think of it this one should just be closed as the fix was included as part of #16.
> 
> I didn't see that. You fooled me with the "refactoring" title :). Can we add a test case for this? Since we already have the Sanitize test case?

I think that case is handled in 1b6a7761ad06cc0bd2dc73aad087b1140a714732 with this test case:

https://github.com/marcosnils/bin/blob/1b6a7761ad06cc0bd2dc73aad087b1140a714732/pkg/assets/assets_test.go#L17

PR #22, opened 2020-07-30
Add HashiCorp releases provider

### LINKED ISSUE #20: Idea: HashiCorp provider
HashiCorp releases many tools that a developer might want to have locally, e.g. `terraform`, `consul`, `vault`, etc.

They have a JSON API: https://releases.hashicorp.com/index.json

So it could be more or less straightforward to build a provider, though maybe it's a little too niche? What do you think?

### LINKED ISSUE #24: Refactor archive recognition
Currently each provider's Fetch function deal with archives on their own, i.e. they test for specific expected archive types which seems redundant.

sirlatrom:
Fixes #20, fixes #24.


sirlatrom:
```bash
bin ls | grep -e Path -e hashi
 Path                                   Version         URL                                                                     Status
 /home/absukl/bin/vault                 1.5.0+ent       releases.hashicorp.com/vault                                            OK
 /home/absukl/bin/consul                1.8.0           releases.hashicorp.com/consul                                           OK
 /home/absukl/bin/envconsul             0.9.3           releases.hashicorp.com/envconsul                                        OK
 /home/absukl/bin/packer                1.6.0           releases.hashicorp.com/packer                                           OK
 /home/absukl/bin/terraform             0.13.0-rc1      releases.hashicorp.com/terraform                                        OK
```

sirlatrom:
May need to rework how the latest version is found as it picks release candidates and enterprise versions.

sirlatrom:
The various equal semvers now result in the user having to break the tie by choosing an edition, e.g. for Vault there's an enterprise one and an enterprise one with HSM which have the same semantic version as the OSS version.

Maybe we could store which edition was chosen at first install, to avoid updates having to ask which edition to get. The semver library considers them equal.

marcosnils:
> Maybe we could store which edition was chosen at first install, to avoid updates having to ask which edition to get. The semver library considers them equal.

SGTM. We can change bin and extend the `version` field stored in the config. We'd have to do something to make it smooth so we don't break people's config. Even though there are not so many bin users yet :)

sirlatrom:
@marcosnils PTAL

pkg/assets/assets.go
```diff
@@ -213,12 +216,165 @@ func ProcessTarGz(r io.Reader) (string, io.Reader, error) {
...
+	var processor processorFunc
+	switch t {
+	case matchers.TypeZip:
+		processor = ProcessZip
+	case matchers.TypeGz:
+		processor = ProcessTarGz
+	case matchers.TypeXz:
+		processor = ProcessTarXz
+	}
+	if processor != nil {
```
sirlatrom, line 361: The idea here is that if the Reader is an archive we recursively check the result (could help split up .tar and .gz/.xz handling in a future commit), and if it's not an archive we return the result.

sirlatrom: I've separated out tar (unarchiving) processing from decompressing (gz, xz, ...) for reuse and decoupling purposes. I've also marked all but assets.ProcessURL as private functions.

sirlatrom: Something that would be nice to achieve is a proper progess bar at The Right Time(tm), i.e. during decompressing or unarchiving, but that's probably not straightforward. Maybe they could be done as separate progress bars, but then again they can happen simultaneously (unarchiving a tar file through the gunzip reader)...

pkg/config/config.go
```diff
@@ -44,9 +44,15 @@ func CheckAndLoad() error {
 	defer f.Close()
 
 	err = json.NewDecoder(f).Decode(&cfg)
+	if err != nil {
```
marcosnils, line 47: Not sure if this will work if starting with an empty file. I have to check it.

pkg/options/options.go
```diff
@@ -11,9 +14,9 @@ func (l LiteralStringer) String() string {
 //Select prompts the user which
 //of the available options is the desired
 //through STDIN and returns the selected one
-func Select(msg string, opts []fmt.Stringer) interface{} {
+func Select(msg string, opts []fmt.Stringer) (interface{}, error) {
```
marcosnils, line 17: Does it make sense for `Select` to return an error if we can't do anything about it? Should't we just os.Exit?

pkg/config/config.go
sirlatrom: I hadn't thought of that. Maybe we should find a way to handle that case more explicitly?

marcosnils: > Maybe we should find a way to handle that case more explicitly?

Yep.. I believe that's better.

pkg/options/options.go
sirlatrom: That will make tests exit rather than error out, I believe. I was running tests on my arm64 device and got an infinite loop since there was no single good match, the test cases expecting the arch to be amd64.

marcosnils: > That will make tests exit rather than error out, I believe. I was running tests on my arm64 device and got an infinite loop since there was no single good match, the test cases expecting the arch to be amd64.

Well.. the idea was to loop infinitely and "force" the user select an option (or press ctrl +c to exit). Maybe I had a typo and selected the wrong number and in that case the program will just exit.

sirlatrom: I can add a check for `io.EOF` which is what `err` becomes in that test situation, but should it then return an error, `nil` or something else?

pkg/config/config.go
sirlatrom: I think this is fixed in 538f9003f03cd1d1ce32665eb4090a1a9ecd3e38.

sirlatrom:
Rebased on the main branch

pkg/assets/assets.go
sirlatrom: Won't fix this in this PR, maybe next time :)

Issue #28, opened 2020-08-10
Print usage when no command is supplied.


arush-sal:
`Bin` should print its usage info when executed without any command.

__Expected behavior__:

```
$ bin
Effortless binary manager

Usage:
  bin [command]

Available Commands:
  help        Help about any command
  install     Installs the specified project
  list        List binaries managed by bin
  prune       Prunes binaries that no longer exist in the system
  remove      Removes binaries managed by bin
  update      Updates one or multiple binaries managed by bin

Flags:
      --debug     Enable debug mode
  -h, --help      help for bin
  -v, --version   version for bin

Use "bin [command] --help" for more information about a command.

```

__Observed Behavior__:

```
$ bin 

 Path   Version         URL     Status
```

__OS/Arch__: Linux/x86
__Bin version__: 
```bin --version
bin version 0.2.0
commit: 1e4d59085ecfa8dd0558d237407382eda76c5d3c
built at: 2020-08-09T15:41:51Z
built by: goreleaser
```

joe733:
> Thx for the contribution. This decision was intentional, since I prefer it this way. https://github.com/marcosnils/bin/pull/35#issuecomment-695840033

Okay, but why then does `bin help` and `bin --help` behave differently?

```console
$ bin --help
Effortless binary manager

Usage:
  bin [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  ensure      Ensures that all binaries listed in the configuration are present
  help        Help about any command
  install     Installs the specified project from a url
  list        List binaries managed by bin
  prune       Prunes binaries that no longer exist in the system
  remove      Removes binaries managed by bin
  update      Updates one or multiple binaries managed by bin

Flags:
      --debug     Enable debug mode
  -h, --help      help for bin
  -v, --version   version for bin

Use "bin [command] --help" for more information about a command.

$ bin help  

 Path				Version		URL					Status
 /home/jcube/.local/bin/pants	v0.5.4		github.com/pantsbuild/scie-pants	OK

$ bin help install

 Path				Version		URL					Status
 /home/jcube/.local/bin/pants	v0.5.4		github.com/pantsbuild/scie-pants	OK

$ bin install --help                                       
Installs the specified project from a url

Usage:
  bin install <url> [flags]

Aliases:
  install, i

Flags:
  -a, --all               Show all possible download options (skip scoring & filtering)
  -f, --force             Force the installation even if the file already exists
  -h, --help              help for install
[...]

PR #35, opened 2020-09-20
Print usage when no command is supplied

szaffarano:


marcosnils:
Thx for the contribution. This decision was intentional, since I prefer it this way.

szaffarano:
Ok, no problem.  I thought that because #28 was already open, it was a valid fix.

Issue #43, opened 2021-02-06
Allow to update binaries using only the binary name


marcosnils:
The way I'm currently updating single binaries is by doing `bin update $(which <program>)`. Even though it's quite nice, I think the UX can be improved by just doing `bin update <program>`. This shouldn't be difficult to do since the [update](https://github.com/marcosnils/bin/blob/master/cmd/update.go#L26) command could perform this check automatically.

cristiand391:
This seems to be resolved in #44.

I would like to be able to remove binaries by name too, any interest in a PR?

marcosnils:
> I would like to be able to remove binaries by name too, any interest in a PR?

Yep, makes sense to allow this as well. Let's make sure to re-use the code from update and delete so the application resolution is the same one.

PR #44, opened 2021-02-06
Update binaries by name

ernesto27:
Hi, marcos.
This is a pull requestes related to the issue https://github.com/marcosnils/bin/issues/43

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/apex/log"
@@ -55,8 +56,14 @@ func newUpdateCmd() *updateCmd {
 
 			// Update single binary
 			if bin != "" {
+				path, errBin := exec.LookPath(bin)
+				if errBin != nil {
+					return printBinaryNotFound(bin)
+				}
+				bin = path
+
 				if b, found := cfg.Bins[bin]; !found {
-					return fmt.Errorf("Binary path %s not found", bin)
+					return printBinaryNotFound(bin)
 
 				} else {
 					if ui, err := getLatestVersion(b); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/update.go
```diff
@@ -55,8 +56,14 @@ func newUpdateCmd() *updateCmd {
 
 			// Update single binary
 			if bin != "" {
+				path, errBin := exec.LookPath(bin)
```
marcosnils, line 59: I think we can be a bit smarter here and avoid the lookup if the user already sends us the path right?  If I do `bin update $(which kind)` like I do today, seems like the lookup is unnecessary. WDYT?

cmd/update.go
```diff
@@ -138,6 +145,10 @@ func newUpdateCmd() *updateCmd {
 	return root
 }
 
+func printBinaryNotFound(bin string) error {
```
marcosnils, line 148: this function doesn't print anything actually. Maybe rename it to "pathBinaryNotFoundError ?

cmd/update.go
ernesto27: Yes, maybe i would add a regexp check for the value / in the path value before call the LookPath function.

But on the other hand i looked at the LookPath docs and i found the is a check of the filestring
"If file contains a slash, it is tried directly and the PATH is not consulted." 
https://golang.org/src/os/exec/lp_unix.go?s=923:965#L24

I'm not sure if this validation of the LookPath function would be enough WDYT?

cmd/update.go
ernesto27: Yes,  that is true ,  no problem with this.

cmd/update.go
marcosnils: > I'm not sure if this validation of the LookPath function would be enough WDYT?

Yes, I saw that as well, but regardless if the argument contains a "/" or not, it still calls the `Stat` syscall afterwards to check if the file exists (https://go.googlesource.com/go/+/go1.15.8/src/os/exec/lp_unix.go#20). I guess this is premature optimization and it's not really necessary. Let's leave it like this for the moment, I don't think we need to change it

ernesto27:
I added a basic check to prevent call LookPath if command is called in this way $(which command),  and change the function name.

I see that other option to  solve this issue would be to use the file HOME/.bin/config.json to check 
the path and prevent to call function LookPath, it would be a better option, WDYT?

marcosnils (approved):

PR #45, opened 2021-02-26
Gitlab generic packages

sirlatrom:
Prefer GitLab Package if found for release version

This allows downloads from private repositories where goreleaser/goreleaser#2002 and goreleaser/goreleaser#2005 are used to upload generic packages for a release.

On private repositories, the uploaded project files made by goreleaser without additional generic uploads are not accessible through the API but only through a web browser session.

### DIFF
--- a/pkg/providers/gitlab.go
+++ b/pkg/providers/gitlab.go
@@ -21,6 +21,7 @@ import (
 type gitLab struct {
 	url    *url.URL
 	client *gitlab.Client
+	token  string
 	owner  string
 	repo   string
 	tag    string
@@ -154,11 +155,11 @@ func (g *gitLab) Fetch() (*File, error) {
 		return nil, err
 	}
 
-	if token, exists := os.LookupEnv("GITLAB_TOKEN"); exists {
+	if g.token != "" {
 		if gf.ExtraHeaders == nil {
 			gf.ExtraHeaders = map[string]string{}
 		}
-		gf.ExtraHeaders["PRIVATE-TOKEN"] = token
+		gf.ExtraHeaders["PRIVATE-TOKEN"] = g.token
 	}
 
 	name, outputFile, err := assets.ProcessURL(gf)
@@ -239,9 +240,14 @@ func newGitLab(u *url.URL) (Provider, error) {
 	}
 
 	token := os.Getenv("GITLAB_TOKEN")
+	hostnameSpecificEnvVarName := fmt.Sprintf("GITLAB_TOKEN_%s", strings.ReplaceAll(u.Hostname(), `.`, "_"))
+	hostnameSpecificToken := os.Getenv(hostnameSpecificEnvVarName)
+	if hostnameSpecificToken != "" {
+		token = hostnameSpecificToken
+	}
 	client, err := gitlab.NewClient(token, gitlab.WithBaseURL(fmt.Sprintf("https://%s/api/v4", u.Hostname())))
 	if err != nil {
 		return nil, err
 	}
-	return &gitLab{url: u, client: client, owner: s[1], repo: s[2], tag: tag}, nil
+	return &gitLab{url: u, client: client, token: token, owner: s[1], repo: s[2], tag: tag}, nil
 }
--- a/pkg/providers/gitlab.go
+++ b/pkg/providers/gitlab.go
@@ -104,7 +104,7 @@ func (g *gitLab) Fetch() (*File, error) {
 		return nil, err
 	}
 	projectUploadsURL := fmt.Sprintf("%s/uploads/", project.WebURL)
-	projectIsPublic := project.Visibility == gitlab.PublicVisibility
+	projectIsPublic := g.token == "" || project.Visibility == gitlab.PublicVisibility
 	log.Debugf("Project is public: %v", projectIsPublic)
 	for _, link := range release.Assets.Links {
 		if projectIsPublic || !strings.HasPrefix(link.URL, projectUploadsURL) {
--- a/pkg/providers/gitlab.go
+++ b/pkg/providers/gitlab.go
@@ -27,11 +27,14 @@ type gitLab struct {
 }
 
 type gitlabFileInfo struct {
-	url, name string
-	score     int
+	url, name, displayName string
+	score                  int
 }
 
 func (g *gitlabFileInfo) String() string {
+	if g.displayName != "" {
+		return g.displayName
+	}
 	return g.name
 }
 
@@ -46,7 +49,7 @@ func (g *gitLab) Fetch() (*File, error) {
 		log.Infof("Getting %s release for %s/%s", g.tag, g.owner, g.repo)
 		release, _, err = g.client.Releases.GetRelease(projectPath, g.tag)
 	} else {
-		//TODO handle case when repo doesn't have releases?
+		// TODO: handle case when repo doesn't have releases?
 		log.Infof("Getting latest release for %s/%s", g.owner, g.repo)
 		name, _, err := g.GetLatestVersion()
 		if err != nil {
@@ -60,6 +63,7 @@ func (g *gitLab) Fetch() (*File, error) {
 	}
 
 	candidates := []*assets.Asset{}
+	candidateURLs := map[string]struct{}{}
 
 	packages, _, err := g.client.Packages.ListProjectPackages(projectPath, &gitlab.ListProjectPackagesOptions{})
 	if err != nil {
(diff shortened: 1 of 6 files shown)


sirlatrom:
@marcosnils What do you think about this approach?

marcosnils:
Will check today Sune :pray:

sent from mobile

Em qui, 4 de mar de 2021 04:53, Sune Keller <notifications@github.com>
escreveu:

pkg/providers/gitlab.go
```diff
@@ -60,25 +60,56 @@ func (g *gitLab) Fetch() (*File, error) {
...
+						v.Version,
+						f.FileName,
+					),
+				})
+			}
+			packageFound = true
+		}
+	}
+
+	if !packageFound {
```
marcosnils, line 92: nit: Do we need the packagesFound bool condition? Can't we just check len(candidates) > 0?

marcosnils (commented):
> the uploaded project files made by goreleaser without additional generic uploads

Is this a `goreleaser` thing? Or still related to the original gitlab issue?

I think the overall PR is ok. However, for public repos with generic packages, there's no way to force `bin` to download Assets if required, right? 

Shall we consider both packages and assets as candidates for public repos?

pkg/providers/gitlab.go
sirlatrom: > Shall we consider both packages and assets as candidates for public repos?

Sounds like a good approach! We could possibly detect the `goreleaser` uploads type of paths combined with the project's visibility setting to determine if the `goreleaser` uploads would be accessible or not. I'll see if it's doable.

sirlatrom: I've attempted the above in d9937e9.

sirlatrom: PS: We could additionally choose to only look for generic packages _if no plausibly publicly available assets are found_.

sirlatrom: Please note that in 028a647, I've altered the asset scoring algorithm so it doesn't give an asset bonus points for mentioning the same desired name components (arch, os, repo name) in _both_ the URL and the name. Previously, such assets might double their score by having those components in both URL and name and would overshadow other candidates. In the case of a GitLab release with both asset URLs and a matching generic package, the generic package would "win" over the asset URL.

pkg/providers/gitlab.go
```diff
@@ -60,8 +64,61 @@ func (g *gitLab) Fetch() (*File, error) {
...
+			}
+		}
+	}
+
+	project, _, err := g.client.Projects.GetProject(projectPath, &gitlab.GetProjectOptions{})
+	if err != nil {
+		return nil, err
+	}
+	projectUploadsURL := fmt.Sprintf("%s/uploads/", project.WebURL)
+	projectIsPublic := project.Visibility == gitlab.PublicVisibility
```
marcosnils, line 107: Hey Sune, I was trying this PR locally in my machine using this repo: gitlab.com/graphviz/graphviz and seems like the `project.Visibility` variable comes empty. However, `project.Public` is set correctly. 

Shall we change this?

sirlatrom: I will test today and see if Public gives the right result for private projects, too. Otherwise I'll make a fallback check so we try both.

sirlatrom: When not authenticated with a personal access token, I don't get the `"visibility"` or `"public"` fields in the API response for a public project. However, with a token towards a public project such as `gitlab.com/graphviz/graphviz`, I get `"visibility": public` but still no `"public"` field, which makes me think it's a deprecated field left around in xanzy/go-gitlab.

So I'm inclined to determine if a project is public like so, since I've now copied the token to the `type gitLab struct`:

```go
projectIsPublic := g.token == "" || project.Visibility == gitlab.PublicVisibility
```

marcosnils: > So I'm inclined to determine if a project is public like so, since I've now copied the token to the `type gitLab struct`:

SGTM!

sirlatrom:
Merge at will?

sirlatrom:
> 
> 
> Hey @sirlatrom I'm doing some preliminary tests and I'm getting some weird results. I'm running this command`./bin --debug install gitlab.com/graphviz/graphviz` I get the following outputs:
> 
> ```
> Multiple matches found, please select one:
> 
>  [1] stable_ubuntu_19.10_debug_x86_64_libgraphviz4-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [2] stable_ubuntu_19.10_debug_x86_64_graphviz-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [3] stable_ubuntu_20.04_debug_x86_64_libgraphviz4-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [4] stable_ubuntu_20.04_debug_x86_64_graphviz-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [5] stable_ubuntu_18.04_debug_x86_64_libgraphviz4-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [6] stable_ubuntu_18.04_debug_x86_64_graphviz-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [7] stable_ubuntu_20.10_debug_x86_64_libgraphviz4-dbgsym_2.46.1-1_amd64.ddeb (asset link)
>  [8] stable_ubuntu_20.10_debug_x86_64_graphviz-dbgsym_2.46.1-1_amd64.ddeb (asset link)
> ```
> 
> This is strange, just `stable_ubuntu` packages and only from the assets link, no generic packages at all.

@marcosnils There are no generic packages because they've chosen not to version their generic packages the same as their tags, e.g. the latest release is v0.0.1110734215291781187721198496291116006578054937833, except from v2.46.1, which looks like the only one. I'll try and change the search order_by to get the largest versions first.

> Another thing that I was thinking that seems the current gitlab provider doesn't support is fectching generic packages from a single release, i.e: `bin install gitlab.com/graphviz/graphviz/-/packages/1325470` and it still outputs the same as the first command.

That's an interesting idea, and I think that should be possible to support as well.

marcosnils:
> I'll try and change the search order_by to get the largest versions first.

SGTM!

marcosnils:
hey @sirlatrom I was checking the last PR against this repository https://gitlab.com/tezos/tezos/-/releases which has both `releases` with the same version number and still I'm not seeing the generic package assets in the results.

```

 [1] x86_64-linux-tezos-binaries.tar.gz (asset link)
 [2] tezos-endorser-alpha (x86_64 Linux) (asset link)
 [3] tezos-baker-alpha (x86_64 Linux) (asset link)
 [4] tezos-accuser-alpha (x86_64 Linux) (asset link)
 [5] tezos-endorser-008-PtEdo2Zk (x86_64 Linux) (asset link)
 [6] tezos-baker-008-PtEdo2Zk (x86_64 Linux) (asset link)
 [7] tezos-accuser-008-PtEdo2Zk (x86_64 Linux) (asset link)
 [8] tezos-endorser-007-PsDELPH1 (x86_64 Linux) (asset link)
 [9] tezos-baker-007-PsDELPH1 (x86_64 Linux) (asset link)
 [10] tezos-accuser-007-PsDELPH1 (x86_64 Linux) (asset link)
 [11] tezos-endorser-006-PsCARTHA (x86_64 Linux) (asset link)
 [12] tezos-baker-006-PsCARTHA (x86_64 Linux) (asset link)
 [13] tezos-accuser-006-PsCARTHA (x86_64 Linux) (asset link)
 [14] tezos-codec (x86_64 Linux) (asset link)
 [15] tezos-signer (x86_64 Linux) (asset link)
 [16] tezos-node (x86_64 Linux) (asset link)
 [17] tezos-client (x86_64 Linux) (asset link)
 [18] tezos-admin-client (x86_64 Linux) (asset link)
```

sirlatrom:
> hey @sirlatrom I was checking the last PR against this repository https://gitlab.com/tezos/tezos/-/releases which has both `releases` with the same version number and still I'm not seeing the generic package assets in the results.
> 
> ```
> Multiple matches found, please select one:
> 
>  [1] x86_64-linux-tezos-binaries.tar.gz (asset link)
> ...
>  [18] tezos-admin-client (x86_64 Linux) (asset link)
> ```

@marcosnils First of all, kudos for finding a GitLab project that actually has both releases and generic packages! I've been searching through public projects for a while without finding one 😄

It looks like there are more fields that are not set when getting a public project through the API without a token, including the `"packages_enabled"` field, which then defaults to `false`. In 234e422 that is handled as a special case.

Through 8670d0c I found out that the generic package files in https://gitlab.com/tezos/tezos/-/releases/8.2.0 score lower than the asset links due to the asset links having "Linux" in the name, which adds some extra score points for the OS, which is missing in the generic package file names.

marcosnils:
> Through [8670d0c](https://github.com/marcosnils/bin/commit/8670d0c41b7c40f3c335400f986f03ae725a6d0a) I found out that the generic package files in https://gitlab.com/tezos/tezos/-/releases/8.2.0 score lower than the asset links due to the asset links having "Linux" in the name, which adds some extra score points for the OS, which is missing in the generic package file names.

Hmm makes sense. I guess we're ok now. 

We can merge this PR and then open issue to support the `install` with a specific package version URL. WDYT?

PR #48, opened 2021-03-05
Remove binaries by name

cristiand391:
Ref. https://github.com/marcosnils/bin/issues/43#issuecomment-791611475

This PR adds support for removing binaries by name.

### DIFF
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/apex/log"
 	"github.com/marcosnils/bin/pkg/config"
@@ -35,13 +34,11 @@ func newRemoveCmd() *removeCmd {
 			existingToRemove := []string{}
 
 			for _, p := range args {
-				if !strings.Contains(p, "/") {
-					var err error
-					p, err = getBinPath(p)
-					if err != nil {
-						return err
-					}
+				p, err := getBinPath(p)
+				if err != nil {
+					return err
 				}
+
 				if _, ok := cfg.Bins[p]; ok {
 					existingToRemove = append(existingToRemove, p)
 					//TODO some providers (like docker) might download
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/apex/log"
 	"github.com/marcosnils/bin/pkg/config"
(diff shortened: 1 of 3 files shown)


cmd/remove.go
```diff
@@ -34,6 +35,16 @@ func newRemoveCmd() *removeCmd {
...
+				// If arg is a binary name, find path.
+				if !strings.Contains(p, "/") {
+					for _, b := range cfg.Bins {
+						if b.RemoteName == p {
+							p = b.Path
+							break
+						}
+					}
+				}
+
```
marcosnils, line 47: Can we move this to a function somewhere and reuse the same logic that's on `update`? Otherwise, if we decide to change the logic eventually, we'd have to do it in two places

cristiand391: At first I tried that approach but since [`config` is private](https://github.com/marcosnils/bin/blob/67ed6c35540ae92cd1dd71a5e960c5831126d781/pkg/config/config.go#L17) I decided to copy/paste from `update`.

Do you think is it a valid use-case to make `config` public?

marcosnils: > Do you think is it a valid use-case to make `config` public?

No need. We can move the logic from update here https://github.com/marcosnils/bin/blob/master/cmd/update.go#L60-L67 to `root.go` so both `update.go` and `remove.go` can actually use it

cristiand391: I'm a little confused, where should I validate args in `root.go`?
I've tried moving it to `Execute()` but it seems it's executed right after the command finishes.

Also, even if I move it to `root.go` it stills needs to get a new `cgf`, in which case I think it would be better to move it to a new file `utils.go` since `root.go` only deals with initial cmd configuration.

marcosnils: > Also, even if I move it to `root.go` it stills needs to get a new `cgf`, in which case I think it would be better to move it to a new file `utils.go` since `root.go` only deals with initial cmd configuration.

What I'm basically saying is to create a private package function somewhere ( I don't really see the need to create a file just for this) which performs this name resolution that's both in `update.go` and `remove.go`. We could create a utils.go file as well, even though I don't usually like those type of filenames since they don't represent anything.

cmd/update.go
```diff
@@ -55,26 +55,19 @@ func newUpdateCmd() *updateCmd {
...
+					var err error
+					bin, err = getBinPath(bin)
+					if err != nil {
+						return err
 					}
 				}
+				b := cfg.Bins[bin]
 
-				if b, found := cfg.Bins[bin]; !found {
-					return fmt.Errorf("Binary path %s not found", bin)
```
cristiand391, line 70: This is now handled by `getBinPath`.

marcosnils: shouldn't `getBinPath` also contains the `if !strings.Contains` condition as well? 

I thought we could move the whole `if` block to that function.

marcosnils (approved):

PR #49, opened 2021-03-15
Adds a "provider" flag to update a install commands

marcosnils:
This flags allows installing binaries from providers that doesn't follow
our initial URL resolution scheme.


cmd/install.go
```diff
@@ -18,19 +18,20 @@ type installCmd struct {
...
 	var cmd = &cobra.Command{
-		Use:           "install",
+		Use:           "install <url>",
 		Aliases:       []string{"i"},
-		Short:         "Installs the specified project",
+		Short:         "Installs the specified project from a url",
 		SilenceUsage:  true,
 		SilenceErrors: true,
-		Args:          cobra.MaximumNArgs(2),
+		Args:          cobra.MinimumNArgs(1),
```
sirlatrom, line 34: I don't believe this change is strictly necessary, as the number of `Args` is not the same as the number of flags. However, if it is needed to make something else work, I don't see a problem with keeping it.

pkg/config/config.go
```diff
@@ -25,6 +25,7 @@ type Binary struct {
 	Version    string `json:"version"`
 	Hash       string `json:"hash"`
 	URL        string `json:"url"`
+	Provider   string `json:"provier"`
```
sirlatrom, line 28:
```suggestion
	Provider   string `json:"provider"`
```

sirlatrom (commented):
A minor typo, otherwise looks good!

cmd/install.go
marcosnils: Oh, this change is unrelated to this PR. I just realized that we didn't configure that correctly and I'm fixing that here since `bin install` always requires a minimum of a 1 argument to work :D

marcosnils:
Thx for reviewing @sirlatrom . I believe it's ready for a second review and merge if approved. :pray:

sirlatrom (approved):
LGTM!

PR #56, opened 2021-03-17
Prevent semantic downgrades in `bin update`

sirlatrom:
Includes:
  - [x] Changes to the `cmd.getLatestVersion()` function
  - [x] Tests of the `cmd.getLatestVersion()` function


cmd/update.go
```diff
@@ -54,28 +55,24 @@ func newUpdateCmd() *updateCmd {
 			cfg := config.Get()
 
 			// Update single binary
+			binsToProcess := cfg.Bins
```
marcosnils, line 58: nit: Can we move this line above the comment. It's a big misleading to understand that the comment is for the `if` statement and not this line :pray:

marcosnils:
LGTM. Added a minor nit comment.

Issue #62, opened 2021-03-19
Add linting


marcosnils:
Bin could use some linting :cupid:

sirlatrom:
Some linting done in in #63

breml:
I recommend [golangci-lint](https://golangci-lint.run/) for linting Go projects.

This linter supports a large set of different linters and one can configure the linting according to the needs (and personal preference). With the default settings, there are currently quite some issues listed. Most of them are related to errors not being checked and unused code.

breml:
And if you want to go more strict on the source code formatting, I recommend `gofumpt` (also part of `gopls`, it can be enabled as default formatting in e.g. vs code Go plugin).

```shell
bin install github.com/mvdan/gofumpt
gofumpt -w -s -l .
```

PR #64, opened 2021-03-19
Add golangci-lint, fix linting issues

breml:
Fixes: #62


pkg/providers/gitlab.go
```diff
@@ -28,20 +28,7 @@ type gitLab struct {
...
-	score                  int
-}
-
-func (g *gitlabFileInfo) String() string {
-	if g.displayName != "" {
-		return g.displayName
-	}
-	return g.name
-}
-
```
sirlatrom, line 42: I imagine these are removed because they are unused?

pkg/providers/gitlab.go
```diff
@@ -53,7 +40,8 @@ func (g *gitLab) Fetch() (*File, error) {
 	} else {
 		// TODO: handle case when repo doesn't have releases?
 		log.Infof("Getting latest release for %s/%s", g.owner, g.repo)
-		name, _, err := g.GetLatestVersion()
+		var name string
+		name, _, err = g.GetLatestVersion()
```
sirlatrom, line 44: I don't understand this change. Is it about re-assigning an existing variable (`err`)?

pkg/providers/hashicorp.go
```diff
@@ -67,21 +67,11 @@ func (g *hashiCorp) listReleases(repoName string) (*hashiCorpRepo, error) {
...
 
-type hashiCorpFileInfo struct {
-	url, name string
-	score     int
-}
-
-func (g *hashiCorpFileInfo) String() string {
-	return g.name
-}
-
```
sirlatrom, line 78: I imagine these are removed because they are unused?

pkg/providers/hashicorp.go
```diff
@@ -90,7 +80,8 @@ func (g *hashiCorp) Fetch() (*File, error) {
 		log.Infof("Getting %s release for %s", g.tag, g.repo)
 		release, err = g.getRelease(g.repo, g.tag)
 	} else {
-		version, _, err := g.GetLatestVersion()
+		var version string
+		version, _, err = g.GetLatestVersion()
```
sirlatrom, line 84: I don't understand the reasoning behind this change. What was the linting error that would recommend this?

sirlatrom (commented):
Thanks for this! I have a few questions as I haven't run the linter yet, and I don't see what caused a few of the changes.

pkg/providers/gitlab.go
breml: Yes, correct. The `unused` linter kicked in here.

pkg/providers/gitlab.go
breml: With `:=` all variables on the left hand side are defined in the respective block, if not yet defined. Therefore, there is a new `err` variable defined. This `err` variable is then re-used on line 48 (`GetRelease`). The check, if the `err` is nil is done outside of the block and therefore checking against the variable defined on line 35 and not against the `err` from the `else` block. It is a combination of shadowing and ineffective assignment.

pkg/providers/hashicorp.go
breml: Yes, correct. The `unused` linter kicked in here.

pkg/providers/hashicorp.go
breml: Same problem as in gitlab provider explained above.

marcosnils (commented):
Shall we add some insturctions on people how to run the linter locally before commit? Otherwise, people will not know and the CI will just complain about it without giving instructions on how to actually run the linter and with what arguments exactly.

sirlatrom:
> 
> 
> Shall we add some insturctions on people how to run the linter locally before commit? Otherwise, people will not know and the CI will just complain about it without giving instructions on how to actually run the linter and with what arguments exactly.

@marcosnils Yes, I'd agree with that. At this stage, maybe the right thing to do would be adding a `CONTRIBUTING.md`?

marcosnils:
hey @sirlatrom I saw you added a pre-commit config file. I guess we need to tell people to run that in their machines before commiting?

sirlatrom:
> hey @sirlatrom I saw you added a pre-commit config file. I guess we need to tell people to run that in their machines before commiting? 

@marcosnils Right, that would make sense!

marcosnils:
> @marcosnils Right, that would make sense!

Since we already merged the CONTRIBUITING guide let's add the instruction there

Issue #57, opened 2021-03-18
Feature Request: Add a possibility to install missing binaries


breml:
I would like to have the possibility to copy the `.bin/config.json` to an other system, download `bin` and run something like `bin ensure` to download and install all the binaries listed in `.bin/config.json` (in the version specified).

Currently, `bin list` is able to detect missing binaries, but from what I can tell, there is no way to fix the *broken* state.

marcosnils:
yes! I had the idea that I already had an issue for this but seems like I never created it. Would you like to give it a try? 
Should be relatively simple to implement.

Issue #50, opened 2021-03-16
GitLab: Support installing based on specific generic package version


sirlatrom:
E.g.

  - https://gitlab.com/tezos/tezos/-/packages/1145959
  - https://gitlab.com/api/v4/projects/3836952/packages/generic/tezos/8.2.0
  - https://gitlab.com/api/v4/projects/3836952/packages/generic/tezos/8.2.0/x86_64-tezos-binaries.tar.gz

https://github.com/marcosnils/bin/issues/45#issuecomment-799904064

sirlatrom:
@breml This issue's scope is currently not to pin versions, but only to allow the first install to be done from a particular package's URL, e.g. if you've been browsing through the `tezos` repo's package list and think that version 8.2.0 sounds good, then you should be able to do `bin install https://gitlab.com/tezos/tezos/-/packages/1145959` and get what you want, just like running `bin install https://github.com/mikefarah/yq/releases/tag/v4.4.0` should give you `yq` version 4.4.0 and not the latest, which is currently v4.6.2.

Version pinning is also something I'm interested in, e.g. for Terraform 0.13.x vs Terraform 0.14.x etc., but should really be discussed in a new issue.

PR #65, opened 2021-03-19
Add ensure command

breml:
Fixes: #57

This is really quick'n'dirty by more or less copy and adopt from the `update` command. But it does work in the sense, that it ensures the latest version of the binaries listed in the config file to be present.

The whole story about version pinning (install the version mentioned in the config file) is not solved. I guess this is addressed in #50.


marcosnils:
LGTM! Can we please change the README to reflect this?

PR #72, opened 2021-03-20
Fall back to repo name if name consists only of replaced literals

### LINKED ISSUE #68: Bug: bin install github.com/Shopify/ejson creates ~/bin/linux
The releases of https://github.com/Shopify/ejson are not optimally named and therefore, bin creates `~/bin/linux` where I would like to get `~/bin/ejson`.

sirlatrom:
Adds additional replacements to match a leading `linux` in the name. This may cause regressions.

Fixes #68.


marcosnils:
Not sure about this yet. Let's keep discussing in #68

Issue #77, opened 2021-03-21
Feature Request: Use `XDG Base Directory Specification` for the configuration directory


rverst:
To avoid cluttering the users $HOME with configuration files/directories I suggest the implementation of the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

marcosnils:
Thx for adding this :pray:

breml:
I like this as well and I look forward to move `~/.bin` to `$HOME/.config/bin`.

PR #78, opened 2021-03-21
Use `XDG Base Directory Specification` for configuration directory

rverst:
Implements #77

Uses the following strategy to get the configuration directory:

- look for `$XDG_CONFIG_HOME` and return `$XDG_CONFIG_HOME/bin` if set and exists
- look for `$HOME/.config` and return `$HOME/config/bin` if it exists
- as default return '$HOME/.bin'

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 
@@ -28,17 +29,16 @@ type Binary struct {
 }
 
 func CheckAndLoad() error {
-	home, err := os.UserHomeDir()
+	configDir, err := getConfigPath()
 	if err != nil {
 		return err
 	}
-
-	configDir := filepath.Join(home, ".bin")
 
 	if err := os.Mkdir(configDir, 0755); err != nil && !os.IsExist(err) {
 		return fmt.Errorf("Error creating config directory [%v]", err)
 	}
 
+	log.Debugf("Config directory is: %s", configDir)
 	f, err := os.OpenFile(filepath.Join(configDir, "config.json"), os.O_RDWR|os.O_CREATE, 0664)
 	if err != nil && !os.IsNotExist(err) {
 		return err
@@ -99,12 +99,12 @@ func RemoveBinaries(paths []string) error {
 }
 
 func write() error {
-	home, err := os.UserHomeDir()
+	configDir, err := getConfigPath()
 	if err != nil {
 		return err
 	}
 
-	f, err := os.OpenFile(filepath.Join(home, ".bin", "config.json"), os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0664)
+	f, err := os.OpenFile(filepath.Join(configDir, "config.json"), os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0664)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


sirlatrom:
This will break existing users' configurations if they have the XDG variable and the existing config file

rverst:
Good point, I've added a check for an existing configuration and return the default directory if a configuration exists.

pkg/config/config.go
```diff
@@ -146,3 +146,36 @@ func GetOS() []string {
...
+	c := os.Getenv("XDG_CONFIG_HOME")
+	if _, err := os.Stat(c); !os.IsNotExist(err) {
+		return path.Join(c, "bin"), nil
+	}
+
+	if homeErr != nil {
+		return "", homeErr
+	}
+	c = path.Join(home, ".config")
+	if _, err := os.Stat(c); !os.IsNotExist(err) {
```
sirlatrom, line 176: What if `err != nil` otherwise?

pkg/config/config.go
```diff
@@ -146,3 +146,36 @@ func GetOS() []string {
...
+func getConfigPath() (string, error) {
+	home, homeErr := os.UserHomeDir()
+	if homeErr == nil {
+		if _, err := os.Stat(path.Join(home, ".bin", "config.json")); !os.IsNotExist(err) {
+			return path.Join(path.Join(home, ".bin")), nil
+		}
+	}
+
+	c := os.Getenv("XDG_CONFIG_HOME")
+	if _, err := os.Stat(c); !os.IsNotExist(err) {
```
sirlatrom, line 168: What if `err != nil` otherwise?

pkg/config/config.go
```diff
@@ -146,3 +146,36 @@ func GetOS() []string {
...
+//     exists and return "$HOME/.bin"
+//   - if "XDG_CONFIG_HOME" is set, return "$XDG_CONFIG_HOME/bin"
+//   - if "$HOME/.config" exists, return "$home/.config/bin"
+//   - default to "$HOME/.bin/"
+// ToDo: move the function to config_unix.go and add a similar function for windows,
+//       %APPDATA% might be the right place on windows
+func getConfigPath() (string, error) {
+	home, homeErr := os.UserHomeDir()
+	if homeErr == nil {
+		if _, err := os.Stat(path.Join(home, ".bin", "config.json")); !os.IsNotExist(err) {
```
sirlatrom, line 162: What if `err != nil` otherwise?

sirlatrom (changes requested):
There may be other error reasons such as permissions errors.

pkg/config/config.go
rverst: If `homeErr != nil` the check for `XDG_HOME_CONFIG` is done first, then the check for `homeErr` returns the error, if there is any.

rverst: Permissions errors will be handled as before in `CheckAndLoad()` or `write()`.

marcosnils (approved):
SGTM. after @sirlatrom approval we can merge.

pkg/config/config.go
breml, line 162: Shouldn't this be `filepath.Join` as well (instead of `path.Join`)?
Also on the next line.

breml:
Thanks for the PR. I like this, I look forward to have this merged.

sirlatrom:
> SGTM. after @sirlatrom approval we can merge.

@marcosnils I'll give it a proper look tomorrow CET, want to make sure I understand the error handling. Or you can merge if you want, I can understand it later :smile:

marcosnils:
> @marcosnils I'll give it a proper look tomorrow CET, want to make sure I understand the error handling. Or you can merge if you want, I can understand it later

No worries, we can wait. There's no need to rush it

PR #87, opened 2021-03-24
Fix bin update on prerelease-only github repos

### LINKED ISSUE #84: Bug: bin is not able to locate releases on Github for bufbuild/buf
I tried to install with the following command:

```shell
$ bin install github.com/bufbuild/buf
   • Getting latest release for bufbuild/buf
   ⨯ command failed            error=GET https://api.github.com/repos/bufbuild/buf/releases/latest: 404 Not Found []
```

But there are clearly releases available on https://github.com/bufbuild/buf/releases
I have not yet investigated, where `bin` is failing.

I am using the latest build:

```shell
$ bin --version
bin version 0.5.0
commit: 7cec71d7c1b77e3db749c010630a8c36122e2801
built at: 2021-03-24T04:11:25Z
built by: goreleaser
```

### LINKED ISSUE #86: Feature Request: support pre releases
Currently pre releases are filtered out by `bin` when checking for releases on Github.

The feature request is to provide an opt-in feature for the user to also use pre releases for `install`, `ensure` and `update`.

Followup to #84.
Releated to #50.

sirlatrom:
Relates to #86.
Fixes #84.

A future PR could add a (per-binary?) option to offer the latest prerelease if newer than latest release.

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -164,6 +164,9 @@ func getLatestVersion(b *config.Binary, p providers.Provider) (*updateInfo, erro
 	if err != nil {
 		return nil, fmt.Errorf("Error checking updates for %s, %w", b.Path, err)
 	}
+	if v == "" && u == "" {
+		return nil, nil
+	}
 
 	if b.Version == v {
 		return nil, nil
--- a/pkg/providers/github.go
+++ b/pkg/providers/github.go
@@ -82,6 +82,20 @@ func (g *gitHub) GetLatestVersion() (string, string, error) {
 	log.Debugf("Getting latest release for %s/%s", g.owner, g.repo)
 	release, _, err := g.client.Repositories.GetLatestRelease(context.TODO(), g.owner, g.repo)
 	if err != nil {
+		// If the error is that no latest release was found, it could be that there are only pre-releases
+		if ghErrResp, ok := err.(*github.ErrorResponse); ok && ghErrResp.Response.StatusCode == http.StatusNotFound {
+			// Get the first release returned by ListReleases
+			releases, _, listErr := g.client.Repositories.ListReleases(context.TODO(), g.owner, g.repo, &github.ListOptions{PerPage: 1})
+			if listErr != nil {
+				return "", "", listErr
+			}
+			if len(releases) > 0 {
+				return releases[0].GetTagName(), releases[0].GetHTMLURL(), nil
+			}
+			// Return original 404/StatusNotFound error from GetLatestRelease
+			return "", "", err
+		}
+
 		return "", "", err
 	}
 
--- a/pkg/providers/github.go
+++ b/pkg/providers/github.go
@@ -29,16 +29,12 @@ func (g *gitHub) Fetch(opts *FetchOpts) (*File, error) {
 
 	// If we have a tag, let's fetch from there
 	var err error
-	var resp *github.Response
 	if len(g.tag) > 0 {
 		log.Infof("Getting %s release for %s/%s", g.tag, g.owner, g.repo)
 		release, _, err = g.client.Repositories.GetReleaseByTag(context.TODO(), g.owner, g.repo, g.tag)
 	} else {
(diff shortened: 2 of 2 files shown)


marcosnils:
hmmmm IIUC this also enables `bin install` to work with pre-releases right?

marcosnils:
If we're doing this, why not going all the way and letting `bin` handle pre-releases on install as well? I guess we need more discussion on #84 before merging this. The first thing that I would have implemented is making `bin update` not to fail, not to fetch the latest pre-release.

Thoughts?

sirlatrom:
> If we're doing this, why not going all the way and letting `bin` handle pre-releases on install as well?

That sounds reasonable, I'll add that in a moment.

marcosnils:
> That sounds reasonable, I'll add that in a moment.

Do we want this though? I've added a comment in #86

marcosnils (approved):

marcosnils:
LGTM. Would like to hear @breml's thoughts before merging

cmd/update.go
```diff
@@ -138,6 +138,9 @@ func getLatestVersion(b *config.Binary, p providers.Provider) (*updateInfo, erro
 	if err != nil {
 		return nil, fmt.Errorf("Error checking updates for %s, %w", b.Path, err)
 	}
+	if v == "" && u == "" {
```
breml, line 141: In which case is it possible, that we do not have an error but still `v` and `u` are empty?

While reviewing this code, it was not helpful, that the two variables are named `v` and `u`, because this does not tell me anything what I can expect to be in these variables. Also the signature of `GetLatestVersion` does not tell me more (just that I get 2 strings and an error as return value). Therefore I propose to use some more speaking variable names.

pkg/providers/github.go
```diff
@@ -69,14 +69,35 @@ func (g *gitHub) Fetch() (*File, error) {
...
+			}
+			if len(releases) > 0 {
+				return releases[0], nil
+			}
+		}
+
+		// Return original 404/StatusNotFound error from GetLatestRelease
+		return nil, err
+	}
+	return release, err
```
breml, line 98: I would explicitly return nil for the error to make it obvious for the reader of the code, that this is the happy path.

```Go
return release, nil
```

pkg/providers/github.go
```diff
@@ -69,14 +69,35 @@ func (g *gitHub) Fetch() (*File, error) {
...
+			releases, _, listErr := g.client.Repositories.ListReleases(ctx, g.owner, g.repo, &github.ListOptions{PerPage: 1})
+			if listErr != nil {
+				return nil, listErr
+			}
+			if len(releases) > 0 {
+				return releases[0], nil
+			}
+		}
+
+		// Return original 404/StatusNotFound error from GetLatestRelease
```
breml, line 95: I think this comment is misleading, because there error could be 404 but it could also be an other error.

pkg/providers/github.go
```diff
@@ -69,14 +69,35 @@ func (g *gitHub) Fetch() (*File, error) {
...
+	release, _, err := g.client.Repositories.GetLatestRelease(ctx, g.owner, g.repo)
+	if err != nil {
+		// If the error is that no latest release was found, it could be that there are only pre-releases
+		if ghErrResp, ok := err.(*github.ErrorResponse); ok && ghErrResp.Response.StatusCode == http.StatusNotFound {
+			// Get the first release returned by ListReleases
+			releases, _, listErr := g.client.Repositories.ListReleases(ctx, g.owner, g.repo, &github.ListOptions{PerPage: 1})
+			if listErr != nil {
+				return nil, listErr
+			}
+			if len(releases) > 0 {
```
breml, line 90: Again, is it possible to not get an error and an empty list of releases? I would expect the API to return an 404 here as well, if there are no releases at all.

marcosnils:
@sirlatrom by any chance do you have any bandwidth to rebase this?

818Nawaf (approved):

818Nawaf (approved):

PR #90, opened 2021-03-27
Add dry-run mode to update command

### LINKED ISSUE #83: Feature Request: bin update check mode
I would like to have a check mode (or dry run mode) for the update command. The command should exit with non 0 exit code, if there are updated version found for any installed binary. If no updates are found, the exit code would be 0 and no output would be printed.

This command I would like to execute on a regular basis (e.g. cron or init of my shell).

Bonus: In order to prevent too frequent requests to e.g. the github API, `bin` could have an option in the config file, how long `bin` does cache the results from this check update command. This would allow to run the check update command on every init of my shell (these can happen quite frequently, whenever I open a new pane/window in tmux).

marcosnils:
Fixes #83

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -15,7 +15,12 @@ import (
 )
 
 type updateCmd struct {
-	cmd *cobra.Command
+	cmd  *cobra.Command
+	opts updateOpts
+}
+
+type updateOpts struct {
+	dryRun bool
 }
 
 type updateInfo struct{ version, url string }
@@ -73,6 +78,8 @@ func newUpdateCmd() *updateCmd {
 			if len(toUpdate) == 0 {
 				log.Infof("All binaries are up to date")
 				return nil
+			} else if root.opts.dryRun {
+				return fmt.Errorf("Command aborted, dry-run mode")
 			}
 
 			// TODO will have to refactor this prompt to a separate function
@@ -129,6 +136,7 @@ func newUpdateCmd() *updateCmd {
 	}
 
 	root.cmd = cmd
+	root.cmd.Flags().BoolVarP(&root.opts.dryRun, "dry-run", "", false, "Only show status, don't prompt for update")
 	return root
 }
 
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -78,8 +78,10 @@ func newUpdateCmd() *updateCmd {
 			if len(toUpdate) == 0 {
 				log.Infof("All binaries are up to date")
 				return nil
-			} else if root.opts.dryRun {
-				return fmt.Errorf("Command aborted, dry-run mode")
+			}
+			
+			if root.opts.dryRun {
+				return fmt.Errorf("Updates found, exit (dry-run mode).")
 			}
 
 			// TODO will have to refactor this prompt to a separate function
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -79,9 +79,9 @@ func newUpdateCmd() *updateCmd {
 				log.Infof("All binaries are up to date")
 				return nil
 			}
-			
+
 			if root.opts.dryRun {
-				return fmt.Errorf("Updates found, exit (dry-run mode).")
+				return wrapErrorWithCode(fmt.Errorf("Updates found, exit (dry-run mode)."), 3, "")
 			}
 
 			// TODO will have to refactor this prompt to a separate function


cmd/update.go
```diff
@@ -73,6 +78,8 @@ func newUpdateCmd() *updateCmd {
 			if len(toUpdate) == 0 {
 				log.Infof("All binaries are up to date")
 				return nil
+			} else if root.opts.dryRun {
+				return fmt.Errorf("Command aborted, dry-run mode")
```
breml, line 82: In my opinion, the else is not necessary here. From a readability point of view, I would write this as follows:

```suggestion
			}
			
			if root.opts.dryRun {
				return fmt.Errorf("Updates found, exit (dry-run mode).")
```

Also I feel, the error message should focus more on the fact, that there are updates, but these are not applied, because of dry run mode.

cmd/update.go
```diff
@@ -129,6 +136,7 @@ func newUpdateCmd() *updateCmd {
 	}
 
 	root.cmd = cmd
+	root.cmd.Flags().BoolVarP(&root.opts.dryRun, "dry-run", "", false, "Only show status, don't prompt for update")
```
breml, line 139: I would mention, that dry run mode exits with non `0` exit code, if updates have been found and `0` otherwise.

```suggestion
	root.cmd.Flags().BoolVarP(&root.opts.dryRun, "dry-run", "", false, "Only show status, don't prompt for update (exit code is non-zero if there are updates available and 0 otherwise)")
```

breml (commented):
LGTM, I have some minor suggestions.

I also lookes at the code in root.go and main.go to understand, how the exit codes are set. Personally, I prefer, if the only place, where `os.Exit` is called, is the `main.main` function. So I would either return an error from `Execute` and check it in `main.main` or return an int (which is the exit code) from `Execute`. That being said, this is purely personal taste and the way the exit func is currently injected does allow for test as well.

cmd/update.go
marcosnils: I actually prefer to read the login in the original way. 

If there `are` updates and we're in dry-run mode, then exit with error. It makes sense to me that are grouped in the same logical block since the condition always need to execute right after checking if there are any updates. 

On the contrary, if we just leave the block as it is here, then `dryRun` condition is completely disconnected to the fact that there are updates to make. 

Thoughts?

breml: My thoughts are as follows:

* I don't agree, that the `if len(toUpdate) == 0 {` is the same logical block as the `if root.opts.dryRun {`. The decides, if there are updates to be made, if not, we can cleanly exit. The second guards the code section, which performs changes to the system from the code section, which does not do any modifications. Imagine, that we could decide to print the number of bytes, that are needed to be downloaded in order to perform the update. We could then implement the dry mode such that we gather this information and print it and only after this, we would interrupt the update.
* In regards to the error message / log output, *Command aborted* to me sounds like it got interrupted unexpectedly. But this is not the case. If I provide the `--dry-run` flag, I actually expect the command to exit. But maybe this is only me (not an native English speaker).

That being said, it is your repo, your tool and your code. If you prefer your version, that is fine for me as well.

marcosnils: > * In regards to the error message / log output, _Command aborted_ to me sounds like it got interrupted unexpectedly. But this is not the case. If I provide the `--dry-run` flag, I actually expect the command to exit. But maybe this is only me (not an native English speaker).

I think error message changes are ok. I didn't make any thoughts on these.

cmd/update.go
marcosnils: I don't think this clarification belongs here. It should be on a `man` page at most, but I don't usually see CLI flags saying what status code they return.

cmd/update.go
marcosnils: > * I don't agree, that the `if len(toUpdate) == 0 {` is the same logical block as the `if root.opts.dryRun {`. The decides, if there are updates to be made, if not, we can cleanly exit. The second guards the code section, which performs changes to the system from the code section, which does not do any modifications. Imagine, that we could decide to print the number of bytes, that are needed to be downloaded in order to perform the update. We could then implement the dry mode such that we gather this information and print it and only after this, we would interrupt the update.

It's still seems like a code smell to me regardless of this example. For me the decision to make it the same block or another one can be easily decided by one simple question. Does the dry-run mode as it is designed right now has any other reason to exist except to exist where there are updates? My answer to that question is "no". Hence, it's clearer to me if those two conditions are in the same scope (block). 

In any case, since  I believe in democracy, it's nice that we're 3 maintainers and @sirlatrom can help us untie this difference of opinions :smile:

cmd/update.go
sirlatrom: If this is meant to be used in scripting I would tend to agree with @breml. It promises the user that they can rely on the exit code. But in that case I would expect a different exit code if there is an actual error.

An alternative to the flag would be a new command, e.g. `updates-available`.

Maybe we could lean on git's porcelain vs. plumbing commands: https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain.

cmd/update.go
sirlatrom: I prefer the separate `if` clause as it more clearly reads as a separate 'point of no return'.

marcosnils: Let's go with separate if's then!

cmd/update.go
marcosnils: > But in that case I would expect a different exit code if there is an actual error.

True, if that's the case we should differentiate from "updates" vs "any other error" 

> An alternative to the flag would be a new command, e.g. `updates-available`.

Hmmm I think at this point I'm more inclined to leave it as a flag than to add a new command just for that. I do agree that we should handle the exit code thing though. I'll update that.

sirlatrom (approved):

PR #92, opened 2021-03-31
Remove unused func param

cristiand391:
Found this one with [unparam](https://github.com/mvdan/unparam)

This PR just remove the unused function parameter and references to it.


pkg/providers/docker.go
```diff
@@ -73,7 +70,7 @@ func newDocker(imageURL string) (Provider, error) {
 // parseImage parses the image returning the repository, tag
 // and an error if it fails. ParseImage handles non-canonical
```
marcosnils, line 71: We need to remove this from the func doc

marcosnils (approved):

Issue #67, opened 2021-03-20
Optimized select for selecting the binary


breml:
When presenting the user a list of potential files to select the correct binary from, the following improvements could be applied to improve the user experience:

* Filter files with extensions, that are most likely not executables or find the files with extensions, that are likely executables (e.g. `.exe`, `.sh` and of course no extension)
* Filter by elements provided in the install url (e.g. repo name in the case of the github provider)
* Evaluate `FileInfo()` information from header of archive (e.g. in tar and zip) to find files with executable bit set
* Order by relevance
* Inverse order, such that the files with the highest relevance are at the end of the list (and therefore closest to the prompt), this is especially important if the archive contains lots of files and the list might span even multiple pages).
* Optional: add a fuzzy search like https://github.com/junegunn/fzf or https://github.com/ktr0731/go-fuzzyfinder

sirlatrom:
The current approach uses a very minimal scoring method, which includes having the repo's name in the binary name or URL's basename. This _should_ already give priority to files that at least include the repo name. However, the other suggestions sound interesting, and especially looking at `FileInfo()` in .tar and .zip files and other archives that support the executable ti sounds like a quick win.

cristiand391:
Finding binaries by reading it's MIME type would be awesome, every time I update a binary I've to manually select it between ~10-20 files.

Is this open to contributions?

marcosnils:
> Is this open to contributions?

Defintely!

For updating binaries I believe there's actually something better we can do. We could save the name of the original file selected the first time and then whenever triggering an update, check if that same file exists to fetch it without asking the user again. Additionally, I suggest we can do what @sirlatrom suggests about improving the scoring method to target the files better to remove altogether the selection step. 

I guess we can implement this in several steps

- Score better archive files based on OS and Arch and don't prompt the user if we have a single high scoring file
- Save the selected file the first time and use the same name upon updates.

akhan4u:
> * Save the selected file the first time and use the same name upon updates.

This will help a lot as it's difficult to check out the same binary when updating. Sometimes I have ended up downloading the checkgen instead of the executable.

sirlatrom:
>
> 
> - Score better archive files based on OS and Arch and don't prompt the user if we have a single high scoring file
> - Save the selected file the first time and use the same name upon updates. 
> 
> 
> 
> 

The first point is already handled for .tar(.*) and .zip archives as the same filtering/selection mechanism is used there as for 'top level' files/assets.

[...]

PR #94, opened 2021-04-10
Don't filter when there's a single asset

### LINKED ISSUE #93: Don't filter assets if there's only one file to download
Until we can finish implementing #67 we need a way to install assets that return a single file which  doesn't have any attributes that increases score. 

i.e `bin install github.com/dapr/cli`

marcosnils:
Fixes #93

Temporary fix until we properly implement #67


sirlatrom (approved):
LGTM

PR #106, opened 2021-07-12
Fix wrapper generated by docker_unix provider

psvo:
Wrapper script generated by docker_unix provider does not contain shebang for the shell. Some shells may fail to execute the script. For example fish is affected.

Add shebang for /bin/sh to make the wrapper script more portable.


marcosnils:
+1. Thx for the contribution!

PR #108, opened 2021-09-26
Allows to set manual download dir if can't be automatically detected

marcosnils:
This commits allows the user to specify a download directorty maunally
if it can't be automatically detected by traversing the PATH folders

Ideally it fixes 107

cc @walidshaari


sirlatrom (approved):
LGTM

PR #116, opened 2021-11-30
Treat empty input as default suggested by prompt

loganlinn:
The prompt,`Do you want to continue? [Y/n]`, uses conventional notation where capitalized letter is default option.
Empty input line results in `Command aborted` which is unexpected behavior to user.
This change treats empty string the same as `"y"`


marcosnils:
> Empty input line results in `Command aborted` which is unexpected behavior to user.

Good catch, thx for the input

Issue #117, opened 2021-12-04
Support bzip2ed binaries like restic


korpa:
From the discussion in #81 

I tried to install restic (https://github.com/restic/restic/releases/tag/v0.12.1) which is a bzip2ed binary. But I get following error:

```shell
$ bin install github.com/restic/restic 
   • Getting latest release for restic/restic
   • Starting download of https://github.com/restic/restic/releases/download/v0.12.1/restic_0.12.1_linux_amd64.bz2
5.92 MiB / 5.92 MiB [------------------------------------------------------------------------------------------------------] 100.00% 102.02 MiB p/s 0s
   ⨯ command failed            error=Error installing binary: open /home/user/bin: file exists
```

So I wonder if bzip2ed binaries are not supported at the moment?

```shell
$ bin --version
bin version 0.10.0
commit: 9aa08f233b1233d649d515a3c147ef6f238003ed
built at: 2021-11-30T18:06:27Z
built by: goreleaser
```

[snip]

marcosnils commented 6 minutes ago

> Just found what's happening. bin has support for .tar.bz2 files, but not just .bz2 files.

> Definitely an "feature" we need to address. Mind adding that please?

korpa:
The issue is within `processBz2`. It always returns an empty name:

```go
func (f *Filter) processBz2(name string, r io.Reader) (string, io.Reader, error) {
	br := bzip2.NewReader(r)

	return "", br, nil
}
```

Compared with `processGz` which returns a name:

```go
// processGz receives a tar.gz file and returns the
// correct file for bin to download
func (f *Filter) processGz(name string, r io.Reader) (string, io.Reader, error) {
	gr, err := gzip.NewReader(r)
	if err != nil {
		return "", nil, err
	}

	return gr.Name, gr, nil
}
```

Problem is now, that bzip2 library is very limited compared to gzip library:

* https://pkg.go.dev/compress/bzip2
* https://pkg.go.dev/compress/gzip

It does not implement the the Header struct. Therefore we don't get back the name without `.bz2` extension.

Just for testing I added following **bad** hack:

```go
func (f *Filter) processBz2(name string, r io.Reader) (string, io.Reader, error) {
	br := bzip2.NewReader(r)

	name = strings.Replace(name, ".bz2", "", -1)

	return name, br, nil
}
```

Voilà, it works:

```shell
go run main.go install github.com/restic/restic
   • Getting latest release for restic/restic
[...]

PR #119, opened 2021-12-05
Fix XZed binaries similar to bzip2ed binaries

korpa:
Fix XZed binaries similar to bzip2ed binaries.

Similar to #117

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -364,7 +364,7 @@ func (f *Filter) processXz(name string, r io.Reader) (string, io.Reader, error) 
 		return "", nil, err
 	}
 
-	return "", xr, nil
+	return name, xr, nil
 }
 
 func (f *Filter) processZip(name string, r io.Reader) (string, io.Reader, error) {


marcosnils:
hey @korpa , how did you test this? Not sure if this is required for `xz` since XZ is different that bz2 in the sense that `xz` does suppor file listing.

korpa:
I haven't tested it in real, but my assumption was that it works similar to bzip2 because I did following in a shell:

```shell
$ touch test-xz.txt
$ xz test-xz.txt
$ ls test-xz.txt.xz
test-xz.txt.xz
$ unxz test-xz.txt.xz
$ ls test-xz.txt
test-xz.txt
```

marcosnils:
> I haven't tested it in real, but my assumption was that it works similar to bzip2 because I did following in a shell:

Hmmm.. intersting.. I see that `xz(1)` has an option `-l` to list information about files that `bzip2` doesn't have. So I thought it could be holding metadata about files but that doesn't seem to be the case.

korpa:
Now I was unsure if xz may be used like bzip2 on the one hand and like zip on the other hand. But checking https://www.kernel.org I say, that the use tar.xz for the kernel (https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.15.6.tar.xz).

Furthermore I check the help of xz:

```
-l, --list          list information about .xz files
```

```shell
$ xz -l test-xz.txt.xz
Strms  Blocks   Compressed Uncompressed  Ratio  Check   Filename
    1       0         32 B          0 B    ---  CRC64   test-xz.txt.xz
```

It really only lists compression infos. So I think my fix is right. But never the less, it would be good to have real world example to test.

Thanks that you already merged it.

PR #123, opened 2021-12-27
Add support to persist package paths

marcosnils:
This commit adds support to persisting packages paths in the bin config
file so that when performing updates, user is not promoted to select the
path again.

Some projects like https://github.com/cli/cli/ add the release version in the archive file which makes this feature useless since it won't bring up any results. For those use-cases, I've added the `-p` flag which skips this path check during updates. 

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>


sirlatrom:
LGTM

PR #130, opened 2022-04-08
Handle case when GitHub repository does not have releases

Stavrospanakakis:
This commit handles the case when a GitHub repository does not have releases. More specifically, it checks if the status code of the HTTP response equals 404 and then it handles the error.


sirlatrom (approved):
~~LGTM~~

I was too slow 🤣

PR #132, opened 2022-04-08
Refactor confirmation prompt to separate pkg

Stavrospanakakis:
This commit refactors the confirmation prompt to a separate pkg.

Signed-off-by: Stavros Panakakis [stavrospanakakis@gmail.com](mailto:stavrospanakakis@gmail.com)

### DIFF
--- a/pkg/prompt/prompt_test.go
+++ b/pkg/prompt/prompt_test.go
@@ -1,0 +1,67 @@
+package prompt
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestConfirm(t *testing.T) {
+	writeToPrompt := func(answer []byte) error {
+		tmpfile, err := ioutil.TempFile("", "tmp")
+		if err != nil {
+			return err
+		}
+
+		defer os.Remove(tmpfile.Name())
+
+		_, err = tmpfile.Write(answer)
+		if err != nil {
+			return err
+		}
+
+		_, err = tmpfile.Seek(0, 0)
+		if err != nil {
+			return err
+		}
+
+		stdin := os.Stdin
+		defer func() {
+			os.Stdin = stdin
+		}()
+
+		os.Stdin = tmpfile
+		err = Confirm("Do you want to continue?")
+		if err != nil {
+			return err
+		}
+
+		if err := tmpfile.Close(); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	t.Run("User confirms that wants to continue", func(t *testing.T) {
+		answers := [][]byte{
+			[]byte("Y\n"),
+			[]byte("y\n"),
+			[]byte("\n"),
+		}
+
+		for _, answer := range answers {
+			err := writeToPrompt(answer)
+			if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/prompt/prompt_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestConfirm(t *testing.T) {
+	writeToPrompt := func(answer []byte) error {
+		tmpfile, err := ioutil.TempFile("", "tmp")
```
marcosnils, line 11: I'm thinking about the reason to have a tmpfile here. Why couldn't we just convert the `answer` to an `io.Reader` and since we're extracting this to a package, we can make the `Confirm` function use the reader as well instead of `os.Stdin`.

thoughts?

Stavrospanakakis: Do you mean something like this?
```go
// prompt.go
package prompt
.
.
.
func Confirm(rd io.Reader, message string) error {
        fmt.Printf("\n%s [Y/n] ", message)
	reader := bufio.NewReader(rd)
        .
        .
}
```
```go
// prune.go
package cmd
.
.
.
func newPruneCmd() *pruneCmd {
        .
        .
        .
        .
	if !root.opts.force {
                err := prompt.Confirm(os.Stdin, "The following paths will be removed. Continue?")
		if err != nil {
		        return err
		}
	}
      .
      .   
```
```go
// prompt_test.go
func TestConfirm(t *testing.T) {
	t.Run("User confirms that wants to continue", func(t *testing.T) {
		answers := [][]byte{
			[]byte("Y\n"),
			[]byte("y\n"),
			[]byte("\n"),
		}

		for _, answer := range answers {
			rd := bytes.NewReader(answer)
			err := Confirm(rd, "Do you want to continue?")
			if err != nil {
				t.Fatal(err)
			}
		}
	})
}
```

marcosnils: sending `os.Stdin` to `Confirm` might be unnecessary. Why not just declaring the `stdin` at the package level and then making the test override it to a custom `io.Reader`?

something like: 

--- main.go ---
``` golang
var stdin io.Reader = os.Stdin

func Confirm(prompt string) error {
 // use stdin here
}
```

--- main_test.go ---
```
func TestConfirm (t *testing.T) {
  stdin = bytes.NewReader([]byte("whatever"))
  // test confirm here

}
```

thoughts?

Stavrospanakakis: Fixed :)

marcosnils: Loved it, thx :heart:

sirlatrom:
How about providing the io.Reader as an argument to Confirm instead?

marcosnils:
> How about providing the io.Reader as a parameter to Confirm instead?

Check https://github.com/marcosnils/bin/pull/132#discussion_r847492299. I though it was easier to handle it separately :pray:

Stavrospanakakis:
The branch had conflicts so I rebased :)

PR #133, opened 2022-04-11
Sort files in archives

### LINKED ISSUE #129: Sort files in archives
Currently, if an archive contains multiple files, they are shown in a random order. It would be better to get a bit more repeatability by sorting them, as I sometimes try to install something, hit 5, it fails (e.g. the file exists), I retry and now I need to hit 8. Not a big deal, just an easy fix.

Stavrospanakakis:
This commit closes #129. When an archive contains multiple files, they are shown in alphabetical order.

Signed-off-by: Stavros Panakakis [stavrospanakakis@gmail.com](mailto:stavrospanakakis@gmail.com)

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -197,7 +197,7 @@ func (f *Filter) FilterAssets(repoName string, as []*Asset) (*FilteredAsset, err
 		}
 
 		sort.SliceStable(generic, func(i, j int) bool {
-			return fmt.Sprint(generic[i]) < fmt.Sprint(generic[j])
+			return generic[i].String() < generic[j].String()
 		})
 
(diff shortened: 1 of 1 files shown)


pkg/assets/assets.go
```diff
@@ -194,6 +195,11 @@ func (f *Filter) FilterAssets(repoName string, as []*Asset) (*FilteredAsset, err
 		for _, f := range matches {
 			generic = append(generic, f)
 		}
+
+		sort.SliceStable(generic, func(i, j int) bool {
+			return fmt.Sprint(generic[i]) < fmt.Sprint(generic[j])
```
marcosnils, line 200: nit:
```suggestion
			return generic[i].String() < generic[j].String()
```

PR #137, opened 2022-05-27
Allow skip errors when upgrading

### LINKED ISSUE #136: Allow skip errors when upgrading
Sometimes, when upgrading there could be situations where users prefer if `bin` skips errors rather than exiting. It'd be nice to have a flag in `bin update` for this.

Any volunteers?

sirlatrom:
Fixes #136.

Adds flag `--continue-on-error` to the `bin update` command, skipping errors with a warning log message if:

1. We get an error while obtaining the latest version metadata, or
2. We get an error while fetching the latest release.

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -20,10 +20,11 @@ type updateCmd struct {
 }
 
 type updateOpts struct {
-	yesToUpdate   bool
-	dryRun        bool
-	all           bool
-	skipPathCheck bool
+	yesToUpdate     bool
+	dryRun          bool
+	all             bool
+	skipPathCheck   bool
+	continueOnError bool
 }
 
 type updateInfo struct{ version, url string }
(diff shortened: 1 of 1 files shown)


sirlatrom:
Real-life example from my own archaic config:

```console
$ bin update
   ⨯ command failed            error=Error checking updates for /home/sune/bin/nerdctl, GET https://api.github.com/repos/AkihiroSuda/nerdctl/releases/latest: 404 Not Found []
```

```console
$ bin update --continue-on-error
   • Error while getting latest version of /home/sune/bin/nerdctl: Error checking updates for /home/sune/bin/nerdctl, GET https://api.github.com/repos/AkihiroSuda/nerdctl/releases/latest: 404 Not Found []
   • All binaries are up to date
```

marcosnils:
@sirlatrom thx for the PR! 

UX feedback: how about if we print a summary of the failed updates at the end with the error description? If I have to go through a long list of updates, and scroll up to see which failed, it might be annoying.

sirlatrom:
> @sirlatrom thx for the PR! 
> 
> UX feedback: how about if we print a summary of the failed updates at the end with the error description? If I have to go through a long list of updates, and scroll up to see which failed, it might be annoying. 

@marcosnils Either that, or the user can try again, at which point they will likely get only warnings and no updates, since those that worked would have already updated?

marcosnils:
> Either that, or the user can try again, at which point they will likely get only warnings and no updates, since those that worked would have already updated?

I've seen people using `bin` in their CI pipeline to install things, so having a good 1st time pass might be worth.

sirlatrom:
> I've seen people using `bin` in their CI pipeline to install things, so having a good 1st time pass might be worth.

I've updated it so that:
* If `-c/--continue-on-error` is set and the `-y/--yes` flag is _not_ set, we'll see the warnings about packages we could not find latest version of _before_ the `[Y/n]` prompt, and any warnings about fetching packages (i.e. we could find latest version but downloading it fails) _after_ fetching those that worked.
* If both `-c/--continue-on-error` and `-y/--yes` are set, we will only see any of those warnings at the end.
* If only `-y/--yes` is set, the command will exit as before on any error, either finding the latest version or downloading the binary.

marcosnils (approved):

PR #142, opened 2022-07-08
add support for ghes

marcosnils:
Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

### DIFF
--- a/pkg/providers/github.go
+++ b/pkg/providers/github.go
@@ -62,7 +62,6 @@ func (g *gitHub) Fetch(opts *FetchOpts) (*File, error) {
 	}
 
 	outFile, err := f.ProcessURL(gf)
-
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


marcosnils:
cc @sandzwerg would you give it a try?

You need to set the `GHES_*` env variables for this to work. Usually the `UPLOAD_URL` and the `BASE_URL` should be the same one, but some GHES deployment use different ones.

pkg/providers/github.go
```diff
@@ -115,12 +114,34 @@ func newGitHub(u *url.URL) (Provider, error) {
...
+
+	if len(gbu) > 0 && len(guu) > 0 && len(gau) > 0 {
+		tc = oauth2.NewClient(context.Background(), oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: token},
+		))
+	} else if token != "" {
 		tc = oauth2.NewClient(context.Background(), oauth2.StaticTokenSource(
 			&oauth2.Token{AccessToken: token},
 		))
 	}
```
sirlatrom, line 133: Why not combine these conditions into one?
```suggestion

	if len(gbu) > 0 && len(guu) > 0 && len(gau) > 0 || token != "" {
		tc = oauth2.NewClient(context.Background(), oauth2.StaticTokenSource(
			&oauth2.Token{AccessToken: token},
		))
	}
```

sirlatrom (changes requested):

pkg/providers/github.go
marcosnils: because I made a mistake and I missed setting the correct token in the `valid` if statement clause :). Fixing

marcosnils:
@sirlatrom mind :eye: one more time? :pray:

sirlatrom (approved):
LGTM 👍

PR #153, opened 2023-01-22
Feature -  Run update command if install url bin already exists

ernesto27:
This PR contains  the following.

If user runs a install command and that binary url exists in config  triggers the update command .

- Update install cmd file
- Add utils file 
- Add tests for new function

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -54,14 +54,16 @@ func newInstallCmd() *installCmd {
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


marcosnils:
One downside of implementing this flow is that `bin install -f $URL` won't work anymore since it'll always try to update now and sometimes people might not want that.

marcosnils:

PR #170, opened 2023-08-07
shorter bin install commands

### LINKED ISSUE #171: shorter bin install command
First of all, thanks for making this awesome tool.

I also noticed many many tools you [just download it and extract it then it works] and there is no need for multiple versions, the latest version is ok enough. so many plugins for asdf or rtx are [redundant](https://github.com/jdxcode/rtx/discussions/686) I think. 
and brew is far from perfect, it's ok on macOS, but not on linux. and original author mxcl is working on tea.xyz now.  With this nice bin tool, it's no need to wget then extract. and no need to install rtx plugins or asdf plugins.

But we can make it better,  like `gh repo clone marcosnils/bin`,  why we need to type github.com. for the releases on GitHub.com can cover 80% of the cases. and been even lazier or there are times when we don't memorize the username. so if there is a default shorthands like [this](https://github.com/jdxcode/rtx/blob/main/src/default_shorthands.rs) could help a lot.

I've draft a simple [pr](#170), with this pr. when I want to install ripgrep. I can just type `bin i rg`

dfang:
close #171

### DIFF
--- a/cmd/default_shorthands.go
+++ b/cmd/default_shorthands.go
@@ -1,0 +1,23 @@
+package cmd
+
+var DEFAULT_SHORTHANDS = make(map[string]string)
+
+func init() {
+	DEFAULT_SHORTHANDS["ripgrep"] = "github.com/BurntSushi/ripgrep"
+	DEFAULT_SHORTHANDS["rg"] = "github.com/BurntSushi/ripgrep"
+	DEFAULT_SHORTHANDS["fzf"] = "github.com/junegunn/fzf"
+	DEFAULT_SHORTHANDS["bin"] = "github.com/marcosnils/bin"
+	DEFAULT_SHORTHANDS["just"] = "github.com/casey/just"
+	DEFAULT_SHORTHANDS["jq"] = "github.com/stedolan/jq"
+	DEFAULT_SHORTHANDS["yq"] = "github.com/mikefarah/yq"
+	DEFAULT_SHORTHANDS["mkcert"] = "github.com/filosottile/mkcert"
+	DEFAULT_SHORTHANDS["golangci-lint"] = "github.com/golangci/golangci-lint"
+	DEFAULT_SHORTHANDS["air"] = "github.com/cosmtrek/air"
+	DEFAULT_SHORTHANDS["fd"] = "github.com/sharkdp/fd"
+	DEFAULT_SHORTHANDS["zoxide"] = "github.com/ajeetdsouza/zoxide"
+	DEFAULT_SHORTHANDS["curlie"] = "github.com/rs/curlie"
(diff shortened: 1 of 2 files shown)


marcosnils:
hey! thx for the contribution! Your PR gave me an idea. How about if we add a new command like `bin alias` to add or remove those aliases that can be stored in bin's config?

We could include the shorthands that you're proposing here as a default during config init or maybe through `bin alias add --default` to populate them. 

Thoughts?

dfang:
`bin alias` is good, but it should not be the top priority thing.  as people can contribute to this file.

I think for the moment smart choose is more important. I mean for this case, 3 should be chosen. I noticed there are issues out on the list.

```
λ bin i dandavison/delta
   • Getting latest release for dandavison/delta
   • Starting download of https://api.github.com/repos/dandavison/delta/releases/assets/111116981
2.53 MiB / 2.53 MiB [---------------------------------------------------------------------------------] 100.00% 122.77 KiB p/s 22s

 [1] delta-0.16.5-x86_64-apple-darwin/LICENSE
 [2] delta-0.16.5-x86_64-apple-darwin/README.md
 [3] delta-0.16.5-x86_64-apple-darwin/delta
 Select an option:
```

marcosnils:
> `bin alias` is good, but it should not be the top priority thing. as people can contribute to this file.

I agree... but I think moving that file shorthands to the `bin` config even without the `bin alias` command would be something very useful as users will be able to add / remove  shorthands by manually editing the config file. WDYT?



> I think for the moment smart choose is more important. I mean for this case, 3 should be chosen. I noticed there are issues out on the list.

yes, agree. I have some ideas on how to improve this, just need to find the time to try them out.

dfang:
> > `bin alias` is good, but it should not be the top priority thing. as people can contribute to this file.
> 
> I agree... but I think moving that file shorthands to the `bin` config even without the `bin alias` command would be something very useful as users will be able to add / remove shorthands by manually editing the config file. WDYT?

it helps ...

marcosnils:
> it helps ...

would you be interested in making this refactor? I can tackle whenever I find some bandwidth though.

dfang:
I think from the perspective of keeping it simple it's not necessary, with this PR, the most widely used tools should be covered, and the else could be installed by `bin i <USER>/<REPO>`. this could make config file simple.

marcosnils:
:+1:. I'll refactor when I find the time. In the meantime I'll hold this PR since I don't want to hardcode custom shorthands in the code that will require a new release every time someone wants to add a new one.

dfang:
Can you enable the discussions feature for this repo?

marcosnils:
Sure, done

dfang:
@marcosnils thanks, https://github.com/marcosnils/bin/discussions/172

marcosnils:

PR #178, opened 2023-10-14
fix: use first github release in the list as latest when latest is not available

matheuscscp:
Hi @marcosnils!

I wasn't being able to install or update `github.com/weaveworks/tf-controller`, so I found this fallback solution for GitHub latest releases which fixes the problem for this repo. The repo does not publish a "latest reference" release, which is what `GetLatestRelease()` looks for. But GitHub has this `ListReleases()` paginated API which lists all the releases starting from the latest (in the common sense meaning of the word, not the "latest reference").

Kind regards,

### DIFF
--- a/pkg/providers/github.go
+++ b/pkg/providers/github.go
@@ -3,6 +3,7 @@ package providers
 import (
 	"context"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


matheuscscp:
oh, sounds like this is a duplicate of #87

marcosnils:

PR #179, opened 2023-11-03
Generic provider

speier:
Started as a Helm provider, trying to turn into a generic provider, see comments below.

~Provider to download Helm releases from https://get.helm.sh~

~Supports getting latest version from https://get.helm.sh/helm-latest-version or fetching specified version.~

~Unfortunately Helm using a rather strange way for their releases and they don't provide a list of release builds for OS/Arch, so we need to maintain a list of possible architecture and OS versions in the provider.~


marcosnils:
Hey! thx for the PR! retrieving binaries from direct http endpoints has been an idea that I've been waiting to generalize as a provider for a while and seems like this PR could be the perfect segway towards that. WDYT if we take this PR as a starting point and create a new `http` provider that additionally allows receiving a `--lastest-url` flag which basically takes care of this? I'd envision the UX something like 

```shell
bin install --latest-url https://get.helm.sh/helm-latest-version https://get.helm.sh/helm-v3.13.1-linux-arm64.tar.gz
```

The idea is that `bin` will store the `latest-url` config and will use it for future `bin update` runs. 

How does that sound?

speier:
Hi! Yes, sounds good. I was also wondering how to have a generic approach.

What would be nice though to come pre-configured for popular tools maybe, what do you think?

Also do you have a plan on how this config would look like and where to store it?

speier:
Did you have a chance to take a look at the updated PR commits?

MurzNN:
It's a brilliant idea to provide providers with the most popular tools! Always get sad about downloading and updating them manually.

marcosnils:
> Did you have a chance to take a look at the updated PR commits?

hey! my apologies for delaying on this one. Will check it out this week :pray:

cmd/install.go
```diff
@@ -18,9 +18,10 @@ type installCmd struct {
 }
 
 type installOpts struct {
-	force    bool
-	provider string
-	all      bool
+	all       bool
+	force     bool
+	provider  string
+	latestURL string
```
marcosnils, line 24: we should move this `latestURL` inside the generic provider only since it doesn't apply to all of them

cmd/install.go
```diff
@@ -57,7 +58,7 @@ func newInstallCmd() *installCmd {
 			// TODO check if binary already exists in config
 			// and triger the update process if that's the case
 
-			p, err := providers.New(u, root.opts.provider)
+			p, err := providers.New(u, root.opts.provider, root.opts.latestURL)
```
marcosnils, line 61: same as the comment above. The provider itself should be the one who parses the latest url from the provided flags instead of sending this argument to all provider initializations. We'll have to refactor the code a bit but I believe it's worth it. We'll have to use different flagsets (https://pkg.go.dev/flag#NewFlagSet) per provider.

cmd/install.go
```diff
@@ -87,6 +88,7 @@ func newInstallCmd() *installCmd {
 				Hash:        fmt.Sprintf("%x", pResult.Hash.Sum(nil)),
 				URL:         u,
 				Provider:    p.GetID(),
+				LatestURL:   root.opts.latestURL,
```
marcosnils, line 91: Same as other comments, having this in the top level of the config seems weird. I was thinking that we could maybe extend the `provider` key we already have and make it `map` instead of `string`. That way, we can save provider specific configurations in the config. 

We could add a new `GetConfig` methgod to the provider interface for this.

cmd/update.go
```diff
@@ -68,7 +68,7 @@ func newUpdateCmd() *updateCmd {
 			updateFailures := map[*config.Binary]error{}
 
 			for _, b := range binsToProcess {
-				p, err := providers.New(b.URL, b.Provider)
+				p, err := providers.New(b.URL, b.Provider, b.LatestURL)
```
marcosnils, line 71: Same as comments in `install.go`

cmd/update.go
```diff
@@ -108,8 +108,7 @@ func newUpdateCmd() *updateCmd {
 			// the same thing as install logic. Refactor to
 			// use the same code in both places
 			for ui, b := range toUpdate {
-
-				p, err := providers.New(ui.url, b.Provider)
+				p, err := providers.New(ui.url, b.Provider, b.LatestURL)
```
marcosnils, line 111: Same as comments in install.go

cmd/update.go
```diff
@@ -133,6 +132,7 @@ func newUpdateCmd() *updateCmd {
 					Version:     pResult.Version,
 					Hash:        fmt.Sprintf("%x", pResult.Hash.Sum(nil)),
 					URL:         ui.url,
+					LatestURL:   b.LatestURL,
```
marcosnils, line 135: Same as comments in install.go

pkg/config/config.go
```diff
@@ -31,6 +31,7 @@ type Binary struct {
 	Hash       string `json:"hash"`
 	URL        string `json:"url"`
 	Provider   string `json:"provider"`
+	LatestURL  string `json:"latest_url"` // for generic provider
```
marcosnils, line 34: Same as comments in install.go

cmd/ensure.go
```diff
@@ -51,7 +51,7 @@ func newEnsureCmd() *ensureCmd {
 					continue
 				}
 
-				p, err := providers.New(binCfg.URL, binCfg.Provider)
+				p, err := providers.New(binCfg.URL, binCfg.Provider, binCfg.LatestURL)
```
marcosnils, line 54: Same as comments in install.go

cmd/ensure.go
```diff
@@ -71,6 +71,7 @@ func newEnsureCmd() *ensureCmd {
 					Version:    pResult.Version,
 					Hash:       fmt.Sprintf("%x", pResult.Hash.Sum(nil)),
 					URL:        binCfg.URL,
+					LatestURL:  binCfg.LatestURL,
```
marcosnils, line 74: Same as comments in install.go

pkg/providers/generic.go
marcosnils, line 1: nit: renane `generic.go` to `http_generic.go`? WDYT?

pkg/providers/generic.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (g *generic) GetID() string {
+	return "generic"
+}
+
+func (g *generic) GetLatestVersion() (string, string, error) {
+	log.Debugf("Getting latest release for %s", g.name)
+
+	resp, err := http.Get(g.latestURL)
```
marcosnils, line 34: we should check if `latestURL` is not empty here so we can add proper `debug` logs for this with a message like: `latest url not found for $name` or something

pkg/providers/providers.go
```diff
@@ -44,16 +45,16 @@ var (
 	dockerUrlPrefix = regexp.MustCompile("^docker://")
 )
 
-func New(u, provider string) (Provider, error) {
+func New(u, provider, latestURL string) (Provider, error) {
```
marcosnils, line 48: Same as comments in install.go

pkg/providers/providers.go
```diff
@@ -70,5 +71,9 @@ func New(u, provider string) (Provider, error) {
 		return newHashiCorp(purl)
 	}
 
+	if len(latestURL) != 0 && (provider == "generic" || len(provider) == 0) {
```
marcosnils, line 74: Same as comments in install.go

pkg/providers/generic.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return &File{Data: outFile.Source, Name: assets.SanitizeName(outFile.Name, version), Hash: sha256.New(), Version: version}, nil
+}
+
+func (g *generic) fmtNameLink(version string) (string, string) {
+	fname := fmt.Sprintf("%s-%s-%s-%s.%s", g.name, version, g.os, g.arch, g.ext)
+	return fname, fmt.Sprintf("%s/%s", g.baseURL, fname)
+}
+
+func newGeneric(u *url.URL, latestURL string) (Provider, error) {
```
marcosnils, line 90: seems like this provider is not very generic since it expects the URL to have a specific format or otherwise it'll fail. I'll leave a more general comment about this in the issue.

marcosnils:
hey @speier just found the time to start looking at this and it's looking great and in the right direction. The most relevant comment I have is that the provider doesn't seem to be generic. If I understood the code correctly, if you give it an URL with a shape different than https://get.helm.sh/helm-v3.13.1-linux-arm64.tar.gz, it'll just fail to parse it and error out. 

Maybe I didn't explain my idea very well before ( apologies, not a native English speaker). My initial thought was to allow the generic provider to be able to fetch **any** kind of URL regardless of the format. So I could also do something like `bin install https://some_url/some_file.tar.gz`  and it'll also work. Optionally, this provider could receive a `latest-url` argument that it'll use to check if there's a new version (initially, it could only check if the value is different to what's persisted in the config) and download the new file if this is the case. 

LMK if that makes sense or if there's something you see I'm missing. 

Thx again for contributing!

speier:
hi @marcosnils, unfortunately i will be off for the next few weeks, starting from tomorrow.

the explanation for the generic provider is clear, but i've only tested with helm really hence the issue i think. anyhow thanks for the feedback, i can take a look later when i'm back.

speier:
> hey @speier just found the time to start looking at this and it's looking great and in the right direction. The most relevant comment I have is that the provider doesn't seem to be generic. If I understood the code correctly, if you give it an URL with a shape different than https://get.helm.sh/helm-v3.13.1-linux-arm64.tar.gz, it'll just fail to parse it and error out.
> 
> Maybe I didn't explain my idea very well before ( apologies, not a native English speaker). My initial thought was to allow the generic provider to be able to fetch **any** kind of URL regardless of the format. So I could also do something like `bin install https://some_url/some_file.tar.gz` and it'll also work. Optionally, this provider could receive a `latest-url` argument that it'll use to check if there's a new version (initially, it could only check if the value is different to what's persisted in the config) and download the new file if this is the case.
> 
> LMK if that makes sense or if there's something you see I'm missing.
> 
> Thx again for contributing!

Hi! The idea was to use a somewhat standard naming convention: `name-version-os-arch.ext` otherwise not sure how it could work, if the url doesn't contain version/os/arch. Maybe I miss something here, but happy to discuss your ideas. Thx!

818Nawaf (approved):

Issue #177, opened 2023-10-12
Freeze Version


dR3b:
Is it possible to freeze/lock a Version?

dR3b:
@marcosnils Unfortunately, I am not a developer.

marcosnils:
happy to provide some guidance if you want to give it a try. It's never late to give it a try and scratch your own itch :smile:

mbevc1:
I like that idea and it sounds similar to what I was thinking we could add as well; so sometimes upstream would break a release, e.g. missing binaries or just b0rked version and I was thinking we could add an exclude switch to ignore specific updates. Also like the freeze/unfreeze idea as well which would solve this issue too :+1:

PR #180, opened 2023-11-14
add pin/unpin command

### LINKED ISSUE #126: Implement version pinning
It would be great if we could pin versions for apps such as terraform, kubectl or istioctl, where often, it is not desirable to install the latest version.

snwzd:
closes #126

### DIFF
--- a/cmd/pin.go
+++ b/cmd/pin.go
@@ -1,0 +1,50 @@
+package cmd
+
+import (
+	"github.com/marcosnils/bin/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+type pinCmd struct {
+	cmd *cobra.Command
+}
+
+func newPinCmd() *pinCmd {
+	root := &pinCmd{}
+
+	cmd := &cobra.Command{
+		Use:           "pin [<name> | <paths...>]",
+		Short:         "Pins current version of the binaries",
+		SilenceUsage:  true,
+		Args:          cobra.MinimumNArgs(1),
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg := config.Get()
+
+			for _, b := range cfg.Bins {
+				for _, p := range args {
+					if b.RemoteName == p {
+						bin, err := getBinPath(p)
+						if err != nil {
+							return err
+						}
+						updatedCfg := cfg.Bins[bin]
+						updatedCfg.Pin = true
+
+						err = config.UpsertBinary(updatedCfg)
+						if err != nil {
+							return err
+						}
+					}
+
+					// TODO return error for unmatched ones
+				}
+			}
+
+			return nil
+		},
+	}
+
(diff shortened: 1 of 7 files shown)


snwzt:
I think pin/unpin is done for binaries, please tell me if there is anything that I should change.

mbevc1:
Probably related//solving to #177 too :thinking: ?

pkg/config/config.go
```diff
@@ -35,6 +35,7 @@ type Binary struct {
 	// the package path in config so we don't ask the user to select
 	// the path again when upgrading
 	PackagePath string `json:"package_path"`
+	Pin         bool   `json:"pin"`
```
marcosnils, line 38: nit: rename to Pinned?

marcosnils:
@snwzt I was coming back to `bin` after a while and I'd like to merge this. Any chance you could rebase? I'll commit to merge it afterwards!

cmd/pin.go
```diff
@@ -0,0 +1,54 @@
...
+		Short:         "Pins current version of the binaries",
+		SilenceUsage:  true,
+		Args:          cobra.MinimumNArgs(1),
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg := config.Get()
+
+			binaryList := make(map[string]*config.Binary)
+			for _, b := range cfg.Bins {
+				binaryList[b.RemoteName] = b
```
marcosnils, line 30: This should use `getBinPath`  similar to update here https://github.com/marcosnils/bin/blob/91ae3e63bc1e2a499a832ba5f4e44d3c071b1345/cmd/update.go?plain=1#L58 so `pin` can be called either by the binary name or the path.

cmd/unpin.go
```diff
@@ -0,0 +1,54 @@
...
+		Short:         "Unpins current version of the binaries",
+		SilenceUsage:  true,
+		Args:          cobra.MinimumNArgs(1),
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg := config.Get()
+
+			binaryList := make(map[string]*config.Binary)
+			for _, b := range cfg.Bins {
+				binaryList[b.RemoteName] = b
```
marcosnils, line 30: same as: https://github.com/marcosnils/bin/pull/180/files#r1694289553

cmd/update.go
```diff
@@ -59,6 +59,9 @@ func newUpdateCmd() *updateCmd {
 					if err != nil {
 						return err
 					}
+					if cfg.Bins[bin].Pin {
```
marcosnils, line 62: Print an info message here instead of returning an error is better as otherwise, this will interrupt the whole update process.

snwzd:
Hi, I’ve rebased the branch and made the requested changes. Please tell me if there's anything that I should change.

Issue #200, opened 2024-06-15
feat: Remember last release match


mbevc1:
More and more vendors are packaging more files than just a binary, e.g. different packages or SBOMs, and updating is a bit more tedious when trying to pick the desired binary or package from a long list of options :smile: .

Example:
```$ bin update
   • /home/marko/bin/trivy v0.52.1 -> v0.52.2 (https://github.com/aquasecurity/trivy/releases/tag/v0.52.2)

Do you want to continue? [Y/n] 
   • Getting v0.52.2 release for aquasecurity/trivy

Multiple matches found, please select one:

 [1] trivy_0.52.2_Linux-64bit.deb.pem
 [2] trivy_0.52.2_Linux-64bit.deb.sig
 [3] trivy_0.52.2_Linux-64bit.rpm.pem
 [4] trivy_0.52.2_Linux-64bit.rpm.sig
 [5] trivy_0.52.2_Linux-64bit.tar.gz
 [6] trivy_0.52.2_Linux-64bit.tar.gz.pem
 [7] trivy_0.52.2_Linux-64bit.tar.gz.sig
 [8] trivy_0.52.2_Linux-ARM64.deb.pem
 [9] trivy_0.52.2_Linux-ARM64.deb.sig
 [10] trivy_0.52.2_Linux-ARM64.rpm.pem
 [11] trivy_0.52.2_Linux-ARM64.rpm.sig
 [12] trivy_0.52.2_Linux-ARM64.tar.gz
 [13] trivy_0.52.2_Linux-ARM64.tar.gz.pem
 [14] trivy_0.52.2_Linux-ARM64.tar.gz.sig
 [15] trivy_0.52.2_Linux-PPC64LE.deb.pem
 [16] trivy_0.52.2_Linux-PPC64LE.deb.sig
 [17] trivy_0.52.2_Linux-PPC64LE.rpm.pem
 [18] trivy_0.52.2_Linux-PPC64LE.rpm.sig
 [19] trivy_0.52.2_Linux-PPC64LE.tar.gz
 [20] trivy_0.52.2_Linux-PPC64LE.tar.gz.pem
 [21] trivy_0.52.2_Linux-PPC64LE.tar.gz.sig
 Select an option: 5
```

I think it would be great if we'd have a mechanism to remember last match/option and default to that.

PR #208, opened 2024-07-15
Print binary path for asset filtering

sascha-andres:
Sometimes I have to use multiple assets from a release to have that software up and running. For example when using https://github.com/Nukesor/pueue I have to use pueue and pueued.

My issue is, that on update I cannot see whether I have to choose pueued or pueue, often ending up in a state where the server binary is the client binary and vice versa.

Therefore the intention of this PR is to print the binary path to be able to make an informed decision.

### DIFF
--- a/cmd/ensure.go
+++ b/cmd/ensure.go
@@ -76,7 +76,7 @@ func newEnsureCmd() *ensureCmd {
 					return err
 				}
 
-				pResult, err := p.Fetch(&providers.FetchOpts{Version: binCfg.Version})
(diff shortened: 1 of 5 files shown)


marcosnils:
> My issue is, that on update I cannot see whether I have to choose pueued or pueue,

This is correct, but I think this should be solved better instead of prompting the user what to do. For example, if I do `bin update pueue`, `bin` should already know which binary was installed before and shouldn't ask again to pick a new binary. 

This is related to this: https://github.com/marcosnils/bin/issues/200

PR #212, opened 2024-10-17
DOC: there is not option to see (default) settings. Debug does this. …

needleshaped:
…let's add clue to for users. Alternative would be add new option, e.g. 'show config'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ bin list # List current binaries and it's versions
 bin prune # Removes from the DB missing binaries
 bin remove <bin>... # Deletes one or more binaries
 bin update [bin]... # Scans binaries and prompts for update
+bin --debug # show config directory, download path
 ```
 
 ## FAQ
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -77,7 +77,7 @@ func newRootCmd(version string, exit func(int)) *rootCmd {
 		},
 	}
 
-	cmd.PersistentFlags().BoolVar(&root.debug, "debug", false, "Enable debug mode")
+	cmd.PersistentFlags().BoolVar(&root.debug, "debug", false, "Enable debug mode, show configuration")
 	cmd.AddCommand(
 		newInstallCmd().cmd,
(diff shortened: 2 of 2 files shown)


marcosnils (approved):

README.md
```diff
@@ -58,6 +58,7 @@ bin list # List current binaries and it's versions
 bin prune # Removes from the DB missing binaries
 bin remove <bin>... # Deletes one or more binaries
 bin update [bin]... # Scans binaries and prompts for update
+bin --debug # show config directory, download path
```
marcosnils, line 61: I don't think `--debug` should be listed here as an example given that's it's a flag that applies to all commands.

marcosnils (changes requested):

README.md
needleshaped: Your point is valid.

However, what other option do I have to "show config directory, download path"? 
As a new user, I had to google for it, and it was not obvious, maybe on 5th or 10th link I found a suggestion. Thus, I try to reduce learning curve for others.

Alternative would be add new option, e.g. `show config` or `info` or `--info` or add this info to `--version`.

marcosnils: > Alternative would be add new _obvious_ option, e.g. `show config` or `info` or `--info` or add this info to `--version`.

 Adding a `bin config show` doesn't seem that bad. WDYT?

needleshaped: `bin config show`  does sound consistent with many tools out there, and is explicit, obvious option. Would be great!

marcosnils: > `bin config show` does sound consistent with many tools out there, and is explicit, obvious option. Would be great!

 awesome! any chance you're willing to open a PR to address this one?

needleshaped: I really want goes against my not so perfect track record. It's a simple feature, and I should be able to do as my first Go experience. Decided, I'll do it)

marcosnils: Awesome! I'm here to help if you have any questions or need any guidance

PR #232, opened 2025-05-22
Feature: Add `go install`

ahmet2mir:
Hello,

Some tools don't provides release via Github and instead uses go install.
This feature check latest release via proxy golang and install tool using local go cmd.

```
$ ./bin install -p goinstall github.com/jrhouston/tfk8s@v0.1.8
   • Getting v0.1.8 release for github.com/jrhouston/tfk8s
   • Copying for tfk8s@v0.1.8 into /home/ahdemir/bin/tfk8s
   • Done installing tfk8s v0.1.8
 
$ ./bin update tfk8s
   • /home/ahdemir/bin/tfk8s v0.1.8 -> v0.1.10 (github.com/jrhouston/tfk8s)
Do you want to continue? [Y/n] y
   • Getting latest release for github.com/jrhouston/tfk8s
   • Copying for tfk8s@v0.1.10 into /home/ahdemir/bin/tfk8s
   • Done updating /home/ahdemir/bin/tfk8s to v0.1.10

$ ./bin remove tfk8s
```

I also had an issue when using remove, since latest release "existingremove" was removed and bin was unable to remove.

I did 3 commits:

* 1 restore existing remove
* 2 explicit error message when path missing
* 3 add goinstall support

### DIFF
--- a/pkg/providers/goinstall.go
+++ b/pkg/providers/goinstall.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"os/exec"
@@ -44,6 +43,10 @@ func newGoInstall(repo string) (Provider, error) {
 }
 
 func (g *goinstall) Fetch(opts *FetchOpts) (*File, error) {
+	goPath, ok := os.LookupEnv("GOPATH")
+	if !ok {
+		return nil, fmt.Errorf("GOPATH environment variable is mandatory")
+	}
 
 	if (len(g.tag) > 0 && g.tag != "latest") || len(opts.Version) > 0 {
 		if len(opts.Version) > 0 {
@@ -69,11 +72,11 @@ func (g *goinstall) Fetch(opts *FetchOpts) (*File, error) {
 		return nil, fmt.Errorf("failed to install package: %w", err)
 	}
 
-	goBinPath := fmt.Sprintf("%s/bin/%s", os.Getenv("GOPATH"), g.name)
+	goBinPath := filepath.Join(goPath, "bin", g.name)
 
 	file, err := os.Open(os.ExpandEnv(goBinPath))
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to open path: %w", err)
 	}
 	defer file.Close()
 
@@ -83,7 +86,6 @@ func (g *goinstall) Fetch(opts *FetchOpts) (*File, error) {
 		return nil, err
 	}
 
-	// fixme use tmpdir
 	// Clean go file added in gopath file
 	if err := os.Remove(os.ExpandEnv(goBinPath)); err != nil && !os.IsNotExist(err) {
 		return nil, fmt.Errorf("Error removing path %s: %v", os.ExpandEnv(goBinPath), err)
@@ -103,7 +105,7 @@ func (d *goinstall) GetLatestVersion() (string, string, error) {
 	}
 	defer resp.Body.Close()
 
-	body, err := ioutil.ReadAll(resp.Body)
+	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return "", "", err
 	}
--- a/pkg/providers/goinstall.go
+++ b/pkg/providers/goinstall.go
@@ -1,0 +1,126 @@
+package providers
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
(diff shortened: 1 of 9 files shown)


ahmet2mir:
@marcosnils if you could get a look at this MR too :pray:

pkg/providers/goinstall.go
```diff
@@ -0,0 +1,126 @@
...
+	cmd := exec.Command("go", "install", fmt.Sprintf("%s@%s", g.repo, g.tag))
+
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("failed to install package: %w", err)
+	}
+
+	goBinPath := fmt.Sprintf("%s/bin/%s", os.Getenv("GOPATH"), g.name)
```
marcosnils, line 72: `GOPATH` could not be set. I think we should do a bit more checks according to the `go install` help: 

> Executables are installed in the directory named by the GOBIN environment
variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH
environment variable is not set. Executables in $GOROOT
are installed in $GOROOT/bin or $GOTOOLDIR instead of $GOBIN.

pkg/providers/goinstall.go
```diff
@@ -0,0 +1,126 @@
...
+
+	file, err := os.Open(os.ExpandEnv(goBinPath))
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	var buf bytes.Buffer
+	_, err = io.Copy(&buf, file)
```
marcosnils, line 81: do we need to read the whole file in memory here? Why not just assign the `file` to `Data` directly? Since `Data` expects an `io.Reader` which `file` already is.

pkg/providers/providers.go
```diff
@@ -49,14 +51,20 @@ type Provider interface {
...
-	httpUrlPrefix   = regexp.MustCompile("^https?://")
-	dockerUrlPrefix = regexp.MustCompile("^docker://")
+	httpUrlPrefix      = regexp.MustCompile("^https?://")
+	dockerUrlPrefix    = regexp.MustCompile("^docker://")
+	goinstallUrlPrefix = regexp.MustCompile("^goinstall://")
 )
 
 func New(u, provider string) (Provider, error) {
 	if dockerUrlPrefix.MatchString(u) {
+		log.Debugf("Using provider 'docker' for '%s'", u)
```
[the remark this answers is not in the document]
ahmet2mir: I did this way to centralize provider selector and log when bin match a condition.
If i show in the begining i don't know if bin will use docker or github.
When running in debug, you see excatly which line is concerned when you saw the log.

pkg/providers/goinstall.go
ahmet2mir: I use to always close a file when i open it plus after i delete it, so i can't use it in return.

pkg/providers/goinstall.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func newGoInstall(repo string) (Provider, error) {
+	repoUrl := strings.TrimPrefix(repo, "goinstall://")
+	repo, tag, name, latestURL := parseRepo(repoUrl)
+	return &goinstall{repo: repo, tag: tag, name: name, latestURL: latestURL}, nil
+}
+
+func (g *goinstall) Fetch(opts *FetchOpts) (*File, error) {
+	goPath, ok := os.LookupEnv("GOPATH")
```
marcosnils, line 46: how about if we run `go env GOPATH` instead? that should always yield a value given that if `GOPATH` is not set, it'll return `$HOME/go` by default

pkg/providers/goinstall.go
ahmet2mir: Not sure that it will be cross-platform, because Linux systems use reference counting for inodes where Windows lock and prevents deletion.
So to be sure it's cross-platform i kould keep the copy or make a difference using runtime.GOOS, if windows copy else  return file ?

pkg/providers/providers.go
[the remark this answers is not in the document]
ahmet2mir: done.

pkg/providers/goinstall.go
ahmet2mir: done.

ahmet2mir:
> awesome, great! we're almost there. This is the last comment I have: https://github.com/marcosnils/bin/pull/232/files#r2223353884 🙏

done too :)

marcosnils (approved):

PR #236, opened 2025-07-21
docs(readme): split providers with examples

ahmet2mir:
it take me some time to found the right command on some providers, so i split the readme with explicitly showing supported providers with dedicated examples.


ahmet2mir:
will do another MR with other proposal of design

PR #257, opened 2025-12-01
Split config manifest/state and normalize URLs

bresilla:
I use this tool across multiple machines and keep my dotfiles under version control. The problem is that every time I update on one machine and push the dotfiles, the other machines think they already have the latest versions. This happens because the manifest mixes two things that should be separate: the repository list (portable) and the per-machine state (mutable versions).

This PR cleanly separates them. The manifest now contains only the repository names, while the per-machine version state is stored in a separate location. On Linux this follows the XDG spec in ~/.local/share/<app>, and on Windows and macOS it follows their respective standard app-data paths. With this split, one can push git changes to without every update mutating the manifest.

## Summary

- Separate per-machine mutable binary state into its own state file while keeping the manifest minimal and portable.
- Normalize manifest repository URLs to canonical base URLs (e.g., https://github.com/owner/repo) while keeping version-specific release URLs in the state file only.
- Introduce migration logic in CheckAndLoad to automatically split legacy manifests and normalize existing URLs on the first run.

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -89,7 +89,9 @@ func newInstallCmd() *installCmd {
 				URL:         u,
 				Provider:    p.GetID(),
 				PackagePath: pResult.PackagePath,
+				StateURL:    u,
 			})
+
 			if err != nil {
 				return err
 			}
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -140,10 +140,12 @@ func newUpdateCmd() *updateCmd {
 					Path:        b.Path,
 					Version:     pResult.Version,
 					Hash:        fmt.Sprintf("%x", hash),
-					URL:         ui.url,
+					URL:         b.URL,
 					Provider:    p.GetID(),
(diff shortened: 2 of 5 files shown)


Amolith:
Ooh I've been having this same annoyance and just pull `bin` changes from one pretty-much-always-on machine, overwriting local `bin` changes. Looking forward to this in a release once it's ready :heart:

marcosnils:
hey @bresilla completely missed this PR. Are you still in the need for this? happy to review it.

bresilla:
Hey, 

No, no worries. I was actually hoping you'd merge one day. But i do understand, sometimes it's difficult to keep track things here 😊, and life happens... 

I made a hard fork on your repo, and did heavy modifications, so i don't think you'd accept the amount of changes i made. And it's mostly AI generated, but has the features i need 😊 

Thanks again for this great tool, i use it all the times 😊

PR #266, opened 2026-01-26
support sub-path in goinstall:// (implement same behaviour as go buid)

gebi:
Hi,

this PR adds support for sub-path's in `goinstall://` provider as is supported in `go install` itself too.

It implements the same behaviour as go build to find base module path.
Sadly https://pkg.go.dev/golang.org/x/tools/go/vcs is deprecated, thus this is implemented as recommended via call to `go list`.

The only real downside is, that this PR requires at least one additional external call to `go list` for every pkg that is handled via `goinstall://` even if no subpath is actually used (for actual sub-paths it results in one network call per sub-path component on each invocation).

But we get nice sub-path support
```
% ./bin install --debug goinstall://github.com/mgit-at/godeb/cmd/godeb
  • debug logs enabled, version: dev

  • Config directory is: /home/gebi/.config/bin
  • Download path set to /home/gebi/bin
  • Using base module github.com/mgit-at/godeb with sub path "/cmd/godeb"
^^^ new debug output if base module was found
  • Using provider 'goinstall' for 'goinstall://github.com/mgit-at/godeb/cmd/godeb'
  • Getting latest release for github.com/mgit-at/godeb
  • Copying for godeb@v0.0.0-20251128154812-dd468d8cd01d into /home/gebi/bin/godeb
  • Done installing godeb v0.0.0-20251128154812-dd468d8cd01d
```

Without this PR the call results in an error (`go install` directly works though)
```
% bin install --debug goinstall://github.com/mgit-at/godeb/cmd/godeb
  • debug logs enabled, version: dev

  • Config directory is: /home/gebi/.config/bin
  • Download path set to /home/gebi/bin
  • Using provider 'goinstall' for 'goinstall://github.com/mgit-at/godeb/cmd/godeb'
  • Getting latest release for github.com/mgit-at/godeb/cmd/godeb
  ⨯ command failed                                   error=failed to get latest version: invalid character 'o' in literal null (expecting 'u')
% go install github.com/mgit-at/godeb/cmd/godeb@latest
^^^ works, thus this PR
```

## output of list cmd
```
% ./bin list

Path                          Version                             URL                                             Status
/home/gebi/bin/bin            v0.24.2                             goinstall://github.com/marcosnils/bin           OK
/home/gebi/bin/godeb          v0.0.0-20251128154812-dd468d8cd01d  goinstall://github.com/mgit-at/godeb/cmd/godeb  OK
```

## `bin/config.json` looks fine
```json
        "/home/gebi/bin/godeb": {
            "path": "/home/gebi/bin/godeb",
            "remote_name": "godeb",
            "version": "v0.0.0-20251128154812-dd468d8cd01d",
            "hash": "12596ca928ce8ed94d495ae152b4dc60a532ec3b3157c415bb645ebe3e35b473",
            "url": "goinstall://github.com/mgit-at/godeb/cmd/godeb",
            "provider": "goinstall",
            "package_path": "",
            "pinned": false
        },
```

## update / ensure / remove also works
```
% ./bin update --debug
  • debug logs enabled, version: dev

  • Config directory is: /home/gebi/.config/bin
  • Download path set to /home/gebi/bin
  • Using provider 'goinstall' for 'goinstall://github.com/marcosnils/bin'
  • Checking updates for /home/gebi/bin/bin
  • Using base module github.com/mgit-at/godeb with sub path "/cmd/godeb"
  • Using provider 'goinstall' for 'goinstall://github.com/mgit-at/godeb/cmd/godeb'
  • Checking updates for /home/gebi/bin/godeb
...
% ./bin ensure
% ./bin remove godeb
```

### DIFF
--- a/pkg/providers/goinstall.go
+++ b/pkg/providers/goinstall.go
@@ -35,8 +35,14 @@ func parseRepo(path string) (string, string, string, string) {
 	return repo, tag, name, latestURL
 }
 
-func baseModulePath(arg string) (string, bool) {
-	noVer := strings.SplitN(arg, "@", 2)[0]
+func moduleRemoveVersion(mod string) string {
+	if i := strings.LastIndex(mod, "@"); i > -1 {
+		return mod[:i]
+	}
+	return mod
+}
+
+func baseModulePath(noVer string) (string, bool) {
 	parts := strings.Split(noVer, "/")
 	for len(parts) > 0 {
 		mod := strings.Join(parts, "/")
@@ -52,8 +58,9 @@ func baseModulePath(arg string) (string, bool) {
 func newGoInstall(repo string) (Provider, error) {
 	repoUrl := strings.TrimPrefix(repo, "goinstall://")
 	subPath := ""
-	baseRepoUrl, found := baseModulePath(repoUrl)
-	subPath = strings.TrimPrefix(repoUrl, baseRepoUrl)
+	repoUrlNoVer := moduleRemoveVersion(repoUrl)
+	baseRepoUrl, found := baseModulePath(repoUrlNoVer)
+	subPath = strings.TrimPrefix(repoUrlNoVer, baseRepoUrl)
 	if found && subPath != "" {
 		repoUrl = baseRepoUrl
 		log.Debugf("Using base module %s with sub path \"%s\"", repoUrl, subPath)
--- a/pkg/providers/goinstall.go
+++ b/pkg/providers/goinstall.go
@@ -14,7 +14,7 @@ import (
 )
 
 type goinstall struct {
-	name, repo, tag, latestURL string
+	name, repo, subPath, tag, latestURL string
 }
 
 func parseRepo(path string) (string, string, string, string) {
@@ -35,10 +35,31 @@ func parseRepo(path string) (string, string, string, string) {
 	return repo, tag, name, latestURL
 }
 
+func baseModulePath(arg string) (string, bool) {
+	noVer := strings.SplitN(arg, "@", 2)[0]
+	parts := strings.Split(noVer, "/")
+	for len(parts) > 0 {
+		mod := strings.Join(parts, "/")
+		out, err := exec.Command("go", "list", "-m", "-f", "{{.Path}}", mod+"@latest").Output()
+		if err == nil {
+			return strings.TrimSpace(string(out)), true
+		}
+		parts = parts[:len(parts)-1]
+	}
+	return "", false
+}
+
 func newGoInstall(repo string) (Provider, error) {
 	repoUrl := strings.TrimPrefix(repo, "goinstall://")
+	subPath := ""
+	baseRepoUrl, found := baseModulePath(repoUrl)
+	subPath = strings.TrimPrefix(repoUrl, baseRepoUrl)
+	if found && subPath != "" {
+		repoUrl = baseRepoUrl
+		log.Debugf("Using base module %s with sub path \"%s\"", repoUrl, subPath)
+	}
 	repo, tag, name, latestURL := parseRepo(repoUrl)
-	return &goinstall{repo: repo, tag: tag, name: name, latestURL: latestURL}, nil
+	return &goinstall{repo: repo, subPath: subPath, tag: tag, name: name, latestURL: latestURL}, nil
 }
 
 func getGoPath() (string, error) {
@@ -71,7 +92,7 @@ func (g *goinstall) Fetch(opts *FetchOpts) (*File, error) {
 		}
 	}
 
-	cmd := exec.Command("go", "install", fmt.Sprintf("%s@%s", g.repo, g.tag))
+	cmd := exec.Command("go", "install", fmt.Sprintf("%s%s@%s", g.repo, g.subPath, g.tag))
 
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr


pkg/providers/goinstall.go
```diff
@@ -35,10 +35,38 @@ func parseRepo(path string) (string, string, string, string) {
...
+	}
+	return mod
+}
+
+func baseModulePath(noVer string) (string, bool) {
+	parts := strings.Split(noVer, "/")
+	for len(parts) > 0 {
+		mod := strings.Join(parts, "/")
+		out, err := exec.Command("go", "list", "-m", "-f", "{{.Path}}", mod+"@latest").Output()
+		if err == nil {
```
marcosnils, line 50: what happens if `go list` fails for other error than "no matching version". Error handling seems it needs improvement.

pkg/providers/goinstall.go
marcosnils, line 49: I'd really prefer avoiding having to do this. Is this how `go install` actually does it?

pkg/providers/goinstall.go
gebi: you are right, if the PR is "ok" in principle i'd add unit tests too, because the behaviour currently expected by the call to `go list` should be tested too.

pkg/providers/goinstall.go
gebi: i think so, yes...
```
% go install -x github.com/mgit-at/godeb/cmd/godeb@latest
# get https://proxy.golang.org/github.com/@v/list
# get https://proxy.golang.org/github.com/mgit-at/godeb/cmd/@v/list
# get https://proxy.golang.org/github.com/mgit-at/godeb/@v/list
# get https://proxy.golang.org/github.com/mgit-at/@v/list
# get https://proxy.golang.org/github.com/mgit-at/godeb/cmd/godeb/@v/list
# get https://proxy.golang.org/github.com/mgit-at/godeb/@v/list: 200 OK (0.028s)
# get https://proxy.golang.org/github.com/mgit-at/godeb/@latest
# get https://proxy.golang.org/github.com/mgit-at/godeb/cmd/@v/list: 404 Not Found (0.052s)
# get https://proxy.golang.org/github.com/@v/list: 404 Not Found (0.052s)
# get https://proxy.golang.org/github.com/mgit-at/godeb/cmd/godeb/@v/list: 404 Not Found (0.056s)
# get https://proxy.golang.org/github.com/mgit-at/godeb/@latest: 200 OK (0.028s)
# get https://proxy.golang.org/github.com/mgit-at/@v/list: 404 Not Found (0.057s)
WORK=/tmp/go-build3389364809
```

It would also be possible to call `proxy.golang.org` directly and implement the walking via the http api but at the expense of breaking and re-implementing the whole local/private/... module machinery that's hidden beneath `go list`.

sadly i could not find any other way...
the pkg holding `RepoRootForImportPath` as a separate and externally usable implementation for this got **deprecated** because of code rot https://pkg.go.dev/golang.org/x/tools/go/vcs and the only suggested solution is to call `go list` instead.

The real implementation can be found here [vcs.RepoRootForImportPath](https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/vcs/vcs.go?q=symbol%3A%5Cbvcs.RepoRootForImportPath%5Cb%20case%3Ayes) and it's imho too complex and sensitive to extract, as it requires other packages from go/internal too so much so that even the pkg in `x/tools/go` gave up tracking it...

i have a implementation that parses the `-x (trace)` output of `go list`, as this tool leaks the output of `vcs.RepoRootForImportPath` there, which would require no recursive calls but after having done the impl it's that ugly to parse trace output and brittle i looked for other possibilities...

gebi: i think i've an idea how to make the default case quicker and still have a somewhat sane impl, i'll cook something up and push into this PR

marcosnils: > sadly i could not find any other way...
> the pkg holding `RepoRootForImportPath` as a separate and externally usable implementation for this got **deprecated** because of code rot [pkg.go.dev/golang.org/x/tools/go/vcs](https://pkg.go.dev/golang.org/x/tools/go/vcs) and the only suggested solution is to call `go list` instead.

 btw not sure if this helps but we've recently also fixed and extracted this functionality to use it in github.com/dagger/dagger so if needed, it can be used from here: https://github.com/dagger/dagger/blob/ba4c738ed055b37f7191678f43c3436e741c910a/engine/vcs/vcs.go#L342

gebi: nice, i've done something similar, though i extracted my base version from the current state of `go/src/cmd/go/internal/vcs/vcs.go` and the one in dagger seems to be the old deprecated version from `x/tools/go/vcs/vcs.go`.

I'd rather not touch a code that upstream deemed unmaintainable and insecure because it diverted so much that security patches can't be applied.

My idea was to implement `RepoRootForImportPath` via taking `repoRootFromVCSPaths` from current upstream and implement a stub for `repoRootForImportDynamic` using `go list`.

thus we would be shielded from much of the complexity with only the implementation of static matches/hosts (like github/gitlab/...) and one `vcsPing` call for verification from go and everything else goes out of process via `go list`.

current state is, i've a complete extract of up-to-date `vcs.go` running. But after doing that it's imho kinda ugly and defies the purpose of being exactly the same mess and unmaintainable over time, thus my idea from above...

in progress... https://github.com/gebi/goxt

d-enk:
Hi! Same bug here with `goinstall://` paths pointing to a package inside a module (`github.com/googleapis/api-linter/v2/cmd/api-linter`).

An alternative that avoids the `go list` walk-up — it's what [gup](https://github.com/nao1215/gup) does: a `go install`-built binary's build info (`debug/buildinfo.ReadFile`) already records the import path (`info.Path`), the containing module path (`info.Main.Path`) and the installed version (`info.Main.Version`). So:

- **`Fetch`**: just run `go install <path>@latest` — `go` resolves the containing module itself — and read the installed version from the built binary's build info. No pre-resolution needed.
- **`GetLatestVersion`**: take `bi.Main.Path` from the managed binary's build info and run a single `go list -m -f {{.Version}} <modulePath>@latest`. No loop, honours `GOPROXY`. Needs the managed binary's path passed in (e.g. via an optional `BinaryPathSetter` interface in `cmd/update.go`).

Happy to PR this if there's interest.

marcosnils:
hey @d-enk, happy to review and merge a PR if you have the time to  add it

PR #274, opened 2026-03-09
feat: filter releases by arch and remove commonly unwanted assets

aaronflorey:
When installing tools from GitHub releases, the asset picker was sometimes confused by non-binary files bundled in the same release - things like checksums, signatures, README files, and shell completions. On top of that, users on Apple Silicon (arm64) weren't getting aarch64 assets matched, and macOS users were missing releases labelled macos or osx instead of darwin.

This PR fixes all three issues:

- Asset filtering: Metadata files (.sha256, .sig, .asc, etc.) and common archive junk (docs, licenses, completion scripts, man pages) are now automatically excluded before the asset picker runs, both at the release level and when extracting from
  archives like tarballs and zips.
- arm64 support: aarch64 is now recognised as an alias for arm64, so releases targeting Apple Silicon and Linux ARM devices are correctly matched.
- macOS aliases: macos and osx are now treated as aliases for darwin, improving compatibility with tools that don't use Go's OS naming convention in their release filenames.

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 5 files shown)


marcosnils:
woot! sorry I haven't had the time to review. do you still need this?

aaronflorey:
no problem! i made the mistake of PRing from `main`, so i'm preparing a new PR now :)

PR #275, opened 2026-03-18
feat: improve arch/os matching and filter non-installable release files

aaronflorey:
## Summary

Improve asset selection accuracy by expanding OS/arch aliases and filtering out non-installable files (checksums/signatures and archive junk) before scoring/selecting release artifacts.

## What changed

- Added arm64 -> aarch64 alias support in GetArch().
- Added Darwin aliases (macos, osx) in GetOS().
- Expanded SanitizeName() normalization to better strip OS/arch/version patterns across common naming permutations.
- Added HasAnySuffix() helper in pkg/strings.
- Introduced installable-asset filtering to skip metadata artifacts (e.g. .sha256, .asc, checksum files).
- Introduced archive-content filtering to skip non-binary files (README/LICENSE/changelog/docs/images/config samples, completion files, manpages) when selecting binaries from .tar/.zip.
- Added safe fallback behavior: if filtering removes everything, keep original candidates.

## Tests

- Added/expanded tests for:
    - arch alias behavior (pkg/config/config_test.go)
    - sanitize-name handling for Darwin ARM naming variants
    - metadata/junk detection helpers
    - archive filtering behavior, including fallback cases
    - manpage extension detection
- Verified with:
    - go test ./... (pass)

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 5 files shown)


marcosnils:
@aaronflorey another closed one? :thinking: . Do you still wanna get this one in?

PR #276, opened 2026-03-19
Print an error when remove doesn't match any binaries

skorokithakis:
## Summary

- `bin remove` with arguments that don't match any installed binary previously exited silently with no output. This reports unmatched arguments to stderr and returns an error if nothing was removed.
- If some arguments match and others don't, the matched ones are still removed and the unmatched ones are reported.

### DIFF
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -46,7 +46,7 @@ func newRemoveCmd() *removeCmd {
 						// additional things somewhere else, maybe we should
 						// call the provider to do a cleanup here.
 						if err := os.Remove(os.ExpandEnv(bp)); err != nil && !os.IsNotExist(err) {
-							return fmt.Errorf("Error removing path %s: %v", os.ExpandEnv(bp), err)
+							return fmt.Errorf("error removing path %s: %v", os.ExpandEnv(bp), err)
 						}
 						fmt.Fprintf(os.Stderr, "removed binary %s\n", os.ExpandEnv(bp))
 						continue
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -48,6 +48,7 @@ func newRemoveCmd() *removeCmd {
 						if err := os.Remove(os.ExpandEnv(bp)); err != nil && !os.IsNotExist(err) {
 							return fmt.Errorf("Error removing path %s: %v", os.ExpandEnv(bp), err)
 						}
+						fmt.Fprintf(os.Stderr, "removed binary %s\n", os.ExpandEnv(bp))
 						continue
 					}
 				}
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -27,6 +27,7 @@ func newRemoveCmd() *removeCmd {
 			cfg := config.Get()
 
 			existingToRemove := []string{}
+			matched := map[string]bool{}
 
 			for _, b := range cfg.Bins {
 				for _, p := range args {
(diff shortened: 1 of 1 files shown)


marcosnils:
> * `bin remove` with arguments that don't match any installed binary previously exited silently with no output.

 in which case this happens? if I run `bin remove foo` and I don't have `foo` installed on my system, I get an error `executable file not found` and `bin` exits with status code 1

> * If some arguments match and others don't, the matched ones are still removed and the unmatched ones are reported.

yes, this one SGTM!

skorokithakis:
I never get anything :( I always get zero output and I've always been confused by it.

I also tend to pass the Github URL, and that definitely doesn't produce any output (I just tried that). This PR makes it much more clear what's happening, for me.

marcosnils:
> I never get anything :( I always get zero output and I've always been confused by it.

 this is what I get when I'm trying to remove something that doesn't exist
 
 ```
 130|marcos:Projects/bin (master) $ bin remove foo
  ⨯ command failed                                   error=exec: "foo": executable file not found in $PATH
  ```

skorokithakis:
I get this:

```

15:18:54 > ~/Downloads on  master via 🐍 v3.12.3 (.venv) 
❯ bin rm github.com/simulot/immich-go

15:19:09 > ~/Downloads on  master via 🐍 v3.12.3 (.venv) 
❯ 
```

skorokithakis:
Or the similarly confusing:

```

15:19:09 > ~/Downloads on  master via 🐍 v3.12.3 (.venv) 
❯ bin uninstall github.com/tempoxyz/tempo

Path                                                   Version     URL                                                              Status
/home/stavros/Documents/Various/bin/bat                v0.26.1     https://github.com/sharkdp/bat/releases/tag/v0.26.1              OK
/home/stavros/Documents/Various/bin/bin                v0.24.2     https://github.com/marcosnils/bin/releases/tag/v0.24.2           OK
...
```

marcosnils:
@skorokithakis just opened a simpler PR which should address all the issues described here. https://github.com/marcosnils/bin/pull/277. Mind checking if this fixes your situation?

skorokithakis:
It works for "uninstall isn't a valid command" and with the GH URL use case, but:

```
15:33:25 > bin on  remove_skip via 🐹 v1.25.3 
❯ bin
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/marcosnils/bin/cmd.(*rootCmd).Execute(0xc0000129f0, {0xc0000360e0, 0x0, 0x416468?})
	/tmp/bin/cmd/root.go:31 +0x1ba
github.com/marcosnils/bin/cmd.Execute({0xe0adce, 0x3}, 0xe568e0, {0xc0000360e0, 0x0, 0x0})
	/tmp/bin/cmd/root.go:24 +0x79
main.main()
	/tmp/bin/main.go:21 +0x8e
```

skorokithakis:
Also, shouldn't this print something for both?


```
15:34:51 > bin on  remove_skip via 🐹 v1.25.3 
❯ bin rm github.com/simulot/immich-go github.com/simulot/immich-gz
  ⨯ command failed                                   error=exec: "github.com/simulot/immich-go": stat github.com/simulot/immich-go: no such file or directory
```

marcosnils:
> It works for "uninstall isn't a valid command" and with the GH URL use case, but:

sorry, forgot to push one fix.

--- 

FWIW `bin  rm $URL`  is not how `rm` is supposed to be used. `rm` expects either a binary name or path in the filesystem, not a URL

marcosnils:
@skorokithakis mind trying again? I think it should work as intended now.

skorokithakis:
Yep, it works now! It has `ls` as default, which contradicts this line in the PR:

> also removing ls as being the default command as it's confusing

So I'm not sure what the intention is there. However, all the issues I, specifically, had with it are gone, so thank you!

marcosnils:
> > also removing ls as being the default command as it's confusing

this fixes the case where if you run `bin notexistcommand` will fallback to `ls`.  Running `bin` without any arguments should fallback to `ls`, that's intended.

PR #288, opened 2026-05-14
honour HTTP_PROXY/HTTPS_PROXY/NO_PROXY in all HTTP clients

### LINKED ISSUE #287: HTTP/HTTPS proxy not supported
# Issue

`bin` makes outbound HTTP requests in several places but does not honour standard proxy environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`). This makes `bin` unusable in corporate or air-gapped environments that require a proxy to reach the internet.

**Affected code paths:**

- `pkg/assets/assets.go` — uses `http.DefaultClient` directly when downloading assets
- `pkg/providers/goinstall.go` — uses `http.Get` (also backed by `http.DefaultClient`) to query `proxy.golang.org`
- `pkg/providers/hashicorp.go` — passes `http.DefaultClient` directly into the Hashicorp provider
- `pkg/providers/github.go` — constructs an `oauth2` HTTP client without a custom transport, so it too ignores proxy settings

**Expected behaviour:**

All HTTP requests should respect `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` environment variables, consistent with the behaviour of most Go tools (including `go` itself).

marcosnils:
## Summary

Fixes #287 — `bin` was not honouring standard proxy environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`), making it unusable in corporate/air-gapped environments.

- Introduces `pkg/httpclient` with a shared `*http.Client` backed by `http.DefaultTransport` (which calls `http.ProxyFromEnvironment`)
- `pkg/assets/assets.go` — uses `httpclient.Client` instead of `http.DefaultClient`
- `pkg/providers/goinstall.go` — uses `httpclient.Client.Get` instead of `http.Get`
- `pkg/providers/hashicorp.go` — passes `httpclient.Client` instead of `http.DefaultClient`
- `pkg/providers/github.go` — injects `httpclient.Client` via `oauth2.HTTPClient` context key so the oauth2 transport uses a proxy-aware base

## Test plan

- [x] Set `HTTPS_PROXY` to a local proxy (e.g. `mitmproxy`) and run `bin install` for a GitHub, HashiCorp, and `goinstall://` binary — all requests should route through the proxy
- [x] Unset proxy vars and verify normal operation is unchanged
- [x] `make test` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -20,6 +20,7 @@ import (
 	"github.com/h2non/filetype/types"
 	"github.com/krolaw/zipstream"
 	"github.com/marcosnils/bin/pkg/config"
+	"github.com/marcosnils/bin/pkg/httpclient"
 	"github.com/marcosnils/bin/pkg/options"
 	bstrings "github.com/marcosnils/bin/pkg/strings"
 	"github.com/xi2/xz"
@@ -345,7 +346,7 @@ func (f *Filter) ProcessURL(gf *FilteredAsset) (*finalFile, error) {
 		req.Header.Add(name, value)
 	}
 	log.Debugf("Checking binary from %s", gf.URL)
-	res, err := http.DefaultClient.Do(req)
+	res, err := httpclient.Client.Do(req)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


Gorian:
That was really quick - and I tested it works on my end; Thank you! :)

PR #295, opened 2026-07-09
feat: add helm provider

TerryHowe:
## What

Adds a dedicated `helm` provider so `bin install github.com/helm/helm` works.

## Why

Helm does **not** attach binaries to its GitHub releases — the release assets are only `.asc` signatures and checksums. The actual binaries are published on [get.helm.sh](https://get.helm.sh) (e.g. `https://get.helm.sh/helm-v3.16.3-linux-amd64.tar.gz`). As a result the existing GitHub provider finds the release but filters away every asset and fails with "Could not find any compatible files".

This is the same situation the `hashicorp` provider already handles for `releases.hashicorp.com`, so the implementation follows that pattern.

## How

- New `pkg/providers/helm.go`:
  - `GetLatestVersion()` reads `https://get.helm.sh/helm-latest-version`.
  - `Fetch()` builds the candidate download URLs for the platforms Helm publishes and reuses the shared `assets` scoring + archive handling to select and unpack the correct binary (the tarballs contain an `os-arch/helm` entry, which `ProcessURL` unpacks to the executable).
  - Supports pinned versions via release-tag URLs (`github.com/helm/helm/releases/tag/v3.16.3`) and the `ensure` command's `--version`.
- `pkg/providers/providers.go`: route `github.com/helm/helm`, `get.helm.sh`, and `--provider helm` to the new provider (matched before the GitHub branch, with an exact `helm/helm` path check so repos like `helm/helm-mapkubeapis` are unaffected).
- Unit tests in `pkg/providers/helm_test.go` (routing, tag extraction, candidate URLs, version normalization).
- README: new "Helm" section and entry in the sources list.

## Testing

- `go build ./...`, `go vet ./pkg/providers/`, and `go test ./pkg/providers/` all pass.
- End-to-end verified on darwin/arm64:

```
$ bin install github.com/helm/helm ./helm
  • Getting v4.2.2 release for helm
  • Starting download of https://get.helm.sh/helm-v4.2.2-darwin-arm64.tar.gz
  • Done installing helm v4.2.2
$ ./helm version
version.BuildInfo{Version:"v4.2.2", ...}

$ bin install github.com/helm/helm/releases/tag/v3.16.3 ./helm-3163
  • Done installing helm v3.16.3
```

### DIFF
--- a/pkg/providers/providers.go
+++ b/pkg/providers/providers.go
@@ -76,7 +76,7 @@ func New(u, provider string) (Provider, error) {
 	// GitHub branch since github.com/helm/helm would otherwise be captured
 	// there.
 	helmPath := strings.TrimPrefix(purl.Path, "/")
-	if provider == "helm" || purl.Host == "get.helm.sh" || (purl.Host == "github.com" && (helmPath == "helm/helm" || strings.HasPrefix(helmPath, "helm/helm/")) {
+	if provider == "helm" || purl.Host == "get.helm.sh" || (purl.Host == "github.com" && (helmPath == "helm/helm" || strings.HasPrefix(helmPath, "helm/helm/"))) {
 		return newHelm(purl)
 	}
 
--- a/pkg/providers/providers.go
+++ b/pkg/providers/providers.go
@@ -71,15 +71,6 @@ func New(u, provider string) (Provider, error) {
 		return nil, err
 	}
 
-	// Helm publishes its binaries on get.helm.sh rather than as GitHub
-	// release assets, so it needs a dedicated provider. Match it before the
-	// GitHub branch since github.com/helm/helm would otherwise be captured
-	// there.
-	helmPath := strings.TrimPrefix(purl.Path, "/")
-	if provider == "helm" || purl.Host == "get.helm.sh" || (purl.Host == "github.com" && (helmPath == "helm/helm" || strings.HasPrefix(helmPath, "helm/helm/"))) {
-		return newHelm(purl)
-	}
-
 	if strings.Contains(purl.Host, "github") || provider == "github" {
 		return newGitHub(purl)
 	}
@@ -90,6 +81,10 @@ func New(u, provider string) (Provider, error) {
 
 	if strings.Contains(purl.Host, "codeberg") || provider == "codeberg" {
 		return newCodeberg(purl)
+	}
+
+	if strings.Contains(purl.Host, "get.helm.sh") || provider == "helm" {
+		return newHelm(purl)
 	}
 
 	if strings.Contains(purl.Host, "releases.hashicorp.com") || provider == "hashicorp" {
--- a/pkg/providers/providers.go
+++ b/pkg/providers/providers.go
@@ -76,7 +76,7 @@ func New(u, provider string) (Provider, error) {
 	// GitHub branch since github.com/helm/helm would otherwise be captured
 	// there.
 	helmPath := strings.TrimPrefix(purl.Path, "/")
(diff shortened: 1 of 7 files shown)


matheuscscp:
Thanks @TerryHowe! Hey @marcosnils, this is due to https://github.com/helm/helm/issues/32315

marcosnils:
@TerryHowe  thx for sending this over. I'd definitely like to support this. One thing that I'm thinking is if we should unify the hashicorp and the helm providers into a single interface (something like http provider)  and then only override the required "GetLatestVersion" depending on which target you're trying to fetch. 

I'll send an agent to do this and if I like the approach, I'll share it here so we can consolidate.

Thx again for taking the time to improve bin :pray:

marcosnils:
Pushed a refactor commit (8ab1de6) on top of this branch that unifies the new helm provider with the existing hashicorp one, since both fetch artifacts from plain HTTP endpoints with the same flow (resolve version → enumerate candidate assets → score/filter → unpack).

- New `pkg/providers/httprelease.go`: `httpReleaseProvider` implements the `Provider` interface once, on top of a two-method `httpSource` seam (`fetchRelease(version)` and `latestVersion()`). Provider IDs persisted in config are unchanged (`helm` / `hashicorp`).
- `helm.go` and `hashicorp.go` shrink to thin sources: helm keeps the static platform matrix + `helm-latest-version` lookup; hashicorp keeps its `index.json` API and semver selection.
- Behavior changes for hashicorp that fall out of the unification:
  - `Fetch` now passes `PackageName` to the asset filter (matching github/codeberg) and returns `PackagePath`, fixing a latent bug where every hashicorp update wiped the stored package path.
  - The latest-version path no longer fetches the release `index.json` a second time.
  - `newHashiCorp` errors on URLs without a repo path instead of panicking.
- Added httptest-backed hashicorp tests (first coverage for that provider) and `GetLatestVersion`-URL round-trip tests for both sources, since `bin update` re-instantiates providers from that URL.

Verified end-to-end on linux/amd64: latest + pinned helm installs, terraform install, `bin update` (pinned helm v3.16.3 → v4.2.3 via the round-trip URL) and `bin ensure` re-installs at stored versions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

marcosnils:
Sorry, claude pushed this automatically without my review. I'll check it out and come back here asking for a review if I like what it did :pray:

pkg/providers/providers.go
```diff
@@ -71,6 +71,15 @@ func New(u, provider string) (Provider, error) {
 		return nil, err
 	}
 
+	// Helm publishes its binaries on get.helm.sh rather than as GitHub
+	// release assets, so it needs a dedicated provider. Match it before the
+	// GitHub branch since github.com/helm/helm would otherwise be captured
+	// there.
+	helmPath := strings.TrimPrefix(purl.Path, "/")
+	if provider == "helm" || purl.Host == "get.helm.sh" || helmPath == "helm/helm" || strings.HasPrefix(helmPath, "helm/helm/") {
```
matheuscscp: codex/gpt-5.6-sol noticed that when checking the `helmPath` here we could also AND with checking the host to be `github.com`:


```suggestion
	if provider == "helm" || purl.Host == "get.helm.sh" || (purl.Host == "github.com" && (helmPath == "helm/helm" || strings.HasPrefix(helmPath, "helm/helm/")) {
```

matheuscscp (commented):
LGTM! 🚀 Thanks @TerryHowe @marcosnils 

Just one nit

As always, very happy with bin taking care of me 😁

PR #297, opened 2026-07-22
feat: add package update cooldowns

### LINKED ISSUE #284: Feature request: Cooldowns
https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

bin is a package manager without a cooldown. It should support a cooldown, perhaps even a default one of 1 day to avoid supply chain attacks.

mbevc1:
Add a supply-chain hardening feature that delays updating a binary until a configurable cooldown has elapsed since the target version was published, giving the community time to detect malicious releases.

- config: `DefaultCooldown` (global) and per-Binary `Cooldown` fields, plus ParseCooldown/FormatCooldown/EffectiveCooldown/SetDefaultCooldown helpers supporting a friendly duration format (24h, 90m, 7d, 2w).
- providers: extend `GetLatestVersion` to return the release publish time; GitHub/GitLab/Codeberg/goinstall supply real dates, others return zero. Add SupportsPublishDate for static enforceability checks.
- update: skip versions published more recently than the effective cooldown; unknown (zero) publish dates are never gated.
- cmd: new `bin cooldown` command with get/set/unset/list subcommands for the default and per-package cooldowns.
- list: show a `Cooldown` column in `bin ls` (- disabled, n/a unenforceable, value with * for per-package override).
- tests + README docs.

Cooldowns are opt-in (disabled by default) so upgrading never silently blocks updates.

Closes #284

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ scoop install extras/bin
 | `bin pin <binary...>`       | Pin current version (prevent updates)      | `bin pin terraform`              |
 | `bin unpin <binary...>`     | Unpin binaries (allow updates)             | `bin unpin terraform`            |
 | `bin prune`                 | Remove missing binaries from database      | `bin prune`                      |
+| `bin cooldown ...`          | Manage update cooldown periods             | `bin cooldown set 24h`           |
 | `bin help`                  | Show help for any command                  | `bin help install`               |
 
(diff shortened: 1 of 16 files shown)


mbevc1:
@marcosnils how does this look like?

